PR #47, opened 2017-03-16
Add murmur hash support

mredolatti:
This pull request addresses the following issues:
 - Adds support for Murmur3 (32 bits) hash algorithm
 - 
   -  Provides the SDK with two choices: native python and compiled C (considerably faster) libraries.
   -  The user can now include an optional dependency when installing via pip `cpphash` which builds and enables a C++ implementation of this hash function.
 - Adds tests for Murmur3 hash function. Passing in Python2.7 & Python 3.4 for both ASCII & utf8 encoded keys.
 - Fixed some linter errors according to PEP 8 standard

In order to include the C++ implementation of this hash function, a fork was created for this project https://github.com/splitio/mmh3, which was adapted to support Python2 as well and then published to SplitIO's Pypi account.

This can be tested locally using the following commands:

- `virtualenv -p <PYTHON_EXECUTABLE> venv & source venv/bin/activate`
- `git clone https://github.com/splitio/python-client`
- `cd python-client`
- `pip install -e .` (python implmementation of hash) or `pip install -e .[cpphash]` (C++ implementation).
- `python setup.py nosetests`

PYTHON_EXECUTABLE can be used to select which python version (installed on your system) should be used to create the virtualenv where the client will be tested.

Once this is published to PyPI it will be possible to install it using:
`pip install splitio` or
`pip install splitio[redis]` or
`pip install splitio[redis,cpphash]`,
etc...


patricioe:
Thank you.

@mredolatti @sarrubia are we waiting for those folks to accept our PR? can we fork and publish it ourselves?

mredolatti:
Hi @patricioe, we have already forked it into split's github account, https://github.com/splitio/mmh3.
And it's been uploaded to splitio's pypi account as splitmmh3.

PR #57, opened 2017-05-05
Out off the grid - auto refresh split file

mredolatti:
Added an OS-agnostic package to handle FS events in order to detect updates to the split definition file (in LocalhostEnvironmentClient) and trigger a handler that re-parses it, updating the splitFetcher for that client.

- [x] Modifications added
- [x] Tests added


mredolatti:
PR Created incorred from incorrect branch.

Correct one: https://github.com/splitio/python-client/pull/58

PR #64, opened 2017-05-25
Handle case of traffic_allocation being None

mpessas:
In Python 3 you cannot compare None with an integer. As a result, the localhost
implementation fails.


mpessas:
Could you backport this fix in the current stable version (4.x)? Thanks.

sarrubia:
Hey @mpessas We will merge it, run internal testing and released on current version 4.1.x.
I will let you know soon!
Thanks for collaborate with us!

mpessas:
@sarrubia Thanks for the quick response!

PR #68, opened 2017-07-27
[HOTFIX] - reschedule fetcher threads when connection fails

mredolatti:
When an invalid response is received, instead of stopping fetcher threads, reschedule them with a timer


mredolatti:
Merging for hotfix release 5.2.1

PR #84, opened 2018-01-08
remove watchdog and use thread based auto update

mredolatti:
…host client


splitio/brokers.py
```diff
@@ -471,38 +444,11 @@ def get_impression_log(self):
...
-            '''
-            Rebuild split fetcher
-            '''
-            self._client_instance()._split_fetcher = (
-                self._client_instance()._build_split_fetcher()
-            )
+    def refresh_splits(self):
+        while not self._destroyed:
+            time.sleep(self._split_refresh_period)
+            if not self._destroyed:
```
sarrubia: Add comment to not remove this IF NOT

PR #109, opened 2018-11-19
Allow hypens in 'localhost mode' file values.

engstrom:
# Python SDK

## What did you accomplish?
In 'localhost mode' any lines that contain a hypen in the `.split` file are not parsed as expected and an 'Invalid line' warning is logged instead. As far as I can tell hyphens are not illegal characters in feature names or treatment names and should be allowed by the SDK.

This change updates the existing regex used to parse the file to allow hypens in both the feature and treatment names and updates the corresponding test.


SenhorCastor:
Hi Dustin,

Thanks for the PR! We will take a look at this and come back to you

Issue #144, opened 2019-06-11
`xrange` not available in Python3


luisincrespo:
https://github.com/splitio/python-client/blob/9a2f1780da9faa23085ddd5c3c9b4d0c074d1f6a/splitio/tasks/uwsgi_wrappers.py#L160

As you can see in the line above, you're using `xrange`, which is no longer available in Python3.

chillaq:
Hi Luis, thanks for the info, will check internally to see how to fix it and keep compatibility with python 2.x

mredolatti:
@luisincrespo `splitio_client==8.1.1` is out and includes your commit. Thank you very much.
I'll be closing this but don't hesitate to reach out again if you need further assistance or something's not working properly.

Thank you very much again and our apologies for any inconveniences caused.

PR #145, opened 2019-06-11
Use range instead of xrange to fix compatibility issue with Python3.

luisincrespo:
# Python SDK

## Tickets covered:
* Issue: https://github.com/splitio/python-client/issues/144

## What did you accomplish?
* Change usage of `xrange` to `range` to avoid compat issues with Python 3.

## How to test new changes?
* python setup.py test to test everythin
* pytest <SPECIFIC_FILE> to test a file in particular (requires pytest, pytest-cov & pytest-mock to be installed)


mredolatti:
Hi @luisincrespo ,
My apologies for this leftover bug, and thank you very much for submitting a fix. We'll be merging this and shipping it with our next release.
May I know why you're trying to use uWSGI Cache as a synchronization platform? Are you unable to use redis?

luisincrespo:
@mredolatti no problem sounds good!

I was using the uWSGI cache alternative to give it a try since the app is already behind uWSGI, do you recommend using Redis?

mredolatti:
The uWSGI cache is not as performant as redis, due to the lack of data structures, and the need of locking and constant serialization/deserialization of complex data. We offer this alternative only for people working on multi-process applications, where you need a single synchronization point and cannot afford a redis instance within their environment.

uWSGI is in fact a multi-process webserver with great performance, and a huge feature set. We highly recommend it as a web-server but not as a synchronization platform for our sdk.

Ideally you would use uWSGI to serve your application bundled with the Python SDK, and leave the background synchronization process to our  Split Synchronizer (https://help.split.io/hc/en-us/articles/360019686092-Split-synchronizer) connected to the SDK via a Redis instance.
This also allows you to have multiple SDKs (even different languages) use a single redis instance, instead of each SDK connecting to the cloud separately to grab the same information.

Please let me know if you need any help with this approach.

luisincrespo:
hey @mredolatti do you have an ETA for merging this fix? For the time being in the small project I'm working on I'll keep using the uWSGI cache for synchronization. I'm currently installing the package pointing directly to my commit with the fix, but it would be nice if I could just point to your package.

mredolatti:
I'll be adding a couple of unit tests to that module and pushing to staging today. We'll probable release today EOD, or early tomorrow.

Thank you very much!

PR #165, opened 2019-11-17
Feature/add timestamp to track

nhausman1:
# Python SDK

## What did you accomplish?
Currently the timestamp for an event is hardcoded as the current time in the track function. This means that you cannot backdate events if, for example, you wanted to run a script every 24 hours that gathers/calculates relevant events and push them to Split. This backdating feature is however supported in [Split's JSON Rest API](https://docs.split.io/reference#create-events) so it will be great to have similar access through the simplified Python API.

Note that this change is non-breaking as I've implemented the logic such that when timestamp isn't specified the logic falls back to `time.time()`. 

Standard build script/test suite. I've added tests for both positive and negative cases.

### DIFF
--- a/splitio/client/client.py
+++ b/splitio/client/client.py
@@ -366,7 +366,7 @@ class Client(object):  # pylint: disable=too-many-instance-attributes
             self._logger.error('Error recording impressions and metrics')
             self._logger.debug('Error: ', exc_info=True)
 
-    def track(self, key, traffic_type, event_type, value=None, properties=None):
+    def track(self, key, traffic_type, event_type, value=None, properties=None, timestamp=None):
         """
         Track an event.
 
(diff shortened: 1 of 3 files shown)


mredolatti:
Hi @nhausman1 , 
Thank you very much for reaching out with this proposal, I will check with our product side to validate whether they're ok with accepting this functionality, and come back to you.

Regards,
Martin.

mredolatti:
Hi @nhausman1 

While this is something that might be supported on future releases, currently it's still being discussed. I've talked with the SDK PM and we created feature request for this, which would help the product team prioritize it properly. 
Unfortunately, we cannot just include your pull request until we figure out how the API will change, and what the exact functionality will be, since this will affect all SDKs and we need to maintain a certain level of consistency between them.

We're truly sorry for any inconveniences this may cause. If all you need to do is to send previously captured events in batches, we can work with you on a small wrapper for our API. 

We'll keep your request in mind and probably reach out to you looking for feedback when we start working on the final design for this feature.

Thank you very much for understanding.
Regards,
Martin.

chillaq:
Hi @nhausman1, Just want to let you know that our Admin API do support posting bulk events allowing specifying the event timestamp.
Checkout this article for more info: https://help.split.io/hc/en-us/articles/9143779240845-Sending-event-data-using-the-Admin-API

Thanks
Bilal

PR #182, opened 2020-10-02
Make block_until_ready() guarantee factory readiness

sethbrite:
# Python SDK

## What did you accomplish?
The code change introduced in this PR ensures that a factory is ready when `block_until_ready` returns. Prior to this change, the value of `factory.ready` was not guaranteed to be `True` after `block_until_ready()` returns, which is counterintuitive. This behavior was due to the non-deterministic nature of Python thread scheduling.

This change can be tested by running the test suite. I simply removed the `time.sleep(1)` calls that enabled factory readiness assertions to pass before this change. 

## Extra Notes

### DIFF
--- a/splitio/client/factory.py
+++ b/splitio/client/factory.py
@@ -108,16 +108,11 @@ class SplitFactory(object):  # pylint: disable=too-many-instance-attributes
         if self._sdk_ready_flag is not None:
             self._status = Status.NOT_INITIALIZED
             # add a listener that updates the status to READY once the flag is set.
-            ready_updater = threading.Thread(target=self._update_status_when_ready)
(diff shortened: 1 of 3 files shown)


mredolatti:
Hi @sethbrite, thank you very much for raising this and sorry to hear this has been causing trouble. We will take a look and use your PR as a starting point to get this fixed so that no incorrect state is shown after a .block_until_ready() call.

sethbrite:
Hi @mredolatti, thanks! This issue has been causing many intermittent test failures for us. Should this change be accepted, do you have a timeline for its release?

sethbrite:
@mredolatti awesome. Thank you!

mredolatti:
Thank you for reporting this, and once again, sorry for the trouble we caused!

PR #231, opened 2021-04-30
added metadata into streaming client

mmelograno:
# Python SDK

## What did you accomplish?
- Added `metadata` into SSE connection.
- Deprecated `charset` config from redis (use `encoding` instead)

## Extra Notes


mredolatti (approved):
LGTM

PR #260, opened 2022-07-05
Updated Synching Splits to detect if there is a segment in rules

chillaq:
# Python SDK

## What did you accomplish?
 Updated Synching Splits to detect if there is a segment in rules, and sync the segment if it does not exist in storage
Updated tests and ran them locally successfully.
## Extra Notes
Related JIRA: https://splitio.atlassian.net/browse/SDKS-2374


splitio/sync/segment.py
```diff
@@ -145,26 +145,47 @@ def synchronize_segment(self, segment_name, till=None):
...
+
+        :param segment_name: Name of the segment
+        :type segment_name: str
+
+        :return: True if segment exist. False otherwise.
+        :rtype: bool
+        """
+        if self._segment_storage.get(segment_name) != None: 
+            return True
+        else:
```
mmelograno, line 190: the else is not required, only the return

splitio/sync/synchronizer.py
```diff
@@ -233,7 +233,13 @@ def synchronize_splits(self, till):
         """
         _LOGGER.debug('Starting splits synchronization')
         try:
-            self._split_synchronizers.split_sync.synchronize_splits(till)
+            segment_list = self._split_synchronizers.split_sync.synchronize_splits(till, self._split_synchronizers.segment_sync)
+            if segment_list != []:
```
mmelograno, line 237: use len(segment_list) != 0 instead

splitio/sync/split.py
```diff
@@ -1,7 +1,10 @@
 """Splits synchronization logic."""
+from ast import Not
```
mmelograno, line 2: this dependencies are not required

splitio/sync/segment.py
chillaq: else: return False is needed to avoid synching segments if the don't exist in sync/split.py _fetch_until() function

splitio/sync/split.py
chillaq: Fixed

splitio/sync/synchronizer.py
chillaq: Done

splitio/sync/segment.py
```diff
@@ -145,26 +145,45 @@ def synchronize_segment(self, segment_name, till=None):
         attempts = _ON_DEMAND_FETCH_BACKOFF_MAX_RETRIES - remaining_attempts
         if successful_sync:  # succedeed sync
             _LOGGER.debug('Refresh completed in %d attempts.', attempts)
-            return
+            return True
```
mredolatti, line 148: can we add  `:returns:` and `:rtype:` to docstring documenting what this method returns and it's type?

splitio/sync/segment.py
```diff
@@ -145,26 +145,45 @@ def synchronize_segment(self, segment_name, till=None):
...
+        """
+        Check if a segment exists in the storage
+
+        :param segment_name: Name of the segment
+        :type segment_name: str
+
+        :return: True if segment exist. False otherwise.
+        :rtype: bool
+        """
+        if self._segment_storage.get(segment_name) != None:
```
mredolatti, line 187: `return self._segment_storage.get(segment_name) != None`

splitio/sync/split.py
```diff
@@ -42,7 +42,7 @@ def __init__(self, split_api, split_storage):
                                 _ON_DEMAND_FETCH_BACKOFF_BASE,
                                 _ON_DEMAND_FETCH_BACKOFF_MAX_WAIT)
 
-    def _fetch_until(self, fetch_options, till=None):
+    def _fetch_until(self, fetch_options, till=None, segment_sync=None):
```
mredolatti, line 45: segment sync should not be optional since it's used without checks.
still, i'd suggest not passing around the segment sync so deep and just returning a list of all the segments referenced in the splits. later on, when the segments are indeed being fetched, filter out the ones that are already present

splitio/sync/synchronizer.py
```diff
@@ -233,7 +233,13 @@ def synchronize_splits(self, till):
         """
         _LOGGER.debug('Starting splits synchronization')
         try:
-            self._split_synchronizers.split_sync.synchronize_splits(till)
+            segment_list = self._split_synchronizers.split_sync.synchronize_splits(till, self._split_synchronizers.segment_sync)
```
mredolatti, line 236:
```
new_segments = [
    segment if not segment_sync.segment_exist_in_storage(segment)
    for segment in self._split_synchronizers.split_sync.synchronize_splits(till, self._split_synchronizers.segment_sync)
]

File: splitio/sync/split.py [hunk 8]
```diff
@@ -55,32 +55,38 @@ def _fetch_until(self, fetch_options, till=None):
         :return: last change number
         :rtype: int
         """
+        segment_list = []
```
mredolatti, line 58: `referenced_segments = set()` to avoid duplicates

splitio/sync/split.py
```diff
@@ -55,32 +55,38 @@ def _fetch_until(self, fetch_options, till=None):
...
                     self._split_storage.put(splits.from_raw(split))
                 else:
                     self._split_storage.remove(split['name'])
-
+                for segment in self._split_storage.get_segment_names():
+                    _LOGGER.debug('Found segment: %s', segment)
+                    if not segment_sync.segment_exist_in_storage(segment):
+                        _LOGGER.debug('Segment %s does not exist, syncing.', segment)
+                        segment_list.append(segment)
+
```
mredolatti, line 84: i think we're asking for the whole list of segment names in a loop for every updated split.
we can do something along the lines of:

```
           for split in split_changes.get('splits', []):
                if split['status'] == splits.Status.ACTIVE.value:
                    parsed = splits.from_raw(split)
                    self._split_storage.put(parsed)
                    referenced_segments.update(set(parsed.get_segment_names()))
                else:
                    self._split_storage.remove(split['name'])
```

splitio/sync/synchronizer.py
```diff
@@ -233,7 +233,13 @@ def synchronize_splits(self, till):
...
+            if len(segment_list) != 0:
+                success = self._split_synchronizers.segment_sync.synchronize_segments(segment_list)
+                if not success:
+                    _LOGGER.error('Failed to sync segment.')
```
mredolatti, line 240: maybe we can print the list of segments that failed to sync here?

splitio/sync/synchronizer.py
```diff
@@ -248,11 +254,11 @@ def sync_all(self):
                 if not self.synchronize_splits(None):
                     attempts -= 1
                     continue
-
+
```
mredolatti, line 257: you might want to set the editor to automatically remove whitespaces in empty lines on save to avoid these

tests/sync/test_synchronizer.py
```diff
@@ -66,9 +66,50 @@ def run(x, y):
...
+                }]
+            },
+            'partitions': [{
+                'size': 100,
+                'treatment': 'on'
+            }]
+        }]
     }]
 
+    def test_synchronize_splits(self, mocker):
```
mredolatti, line 88: i think it might be worth adding some tests for 
- `_fetch_until` checking that the returned value is correct
- ensuring that `segment_sync` gets called once for every newly referenced segment

splitio/sync/synchronizer.py
```diff
@@ -233,7 +233,13 @@ def synchronize_splits(self, till):
         """
         _LOGGER.debug('Starting splits synchronization')
         try:
-            self._split_synchronizers.split_sync.synchronize_splits(till)
+            segment_list = self._split_synchronizers.split_sync.synchronize_splits(till, self._split_synchronizers.segment_sync)
+            if len(segment_list) != 0:
```
mredolatti, line 237: i think that if the list is not empty we should fetch the new segemnts in bg. we should spawn a thread, set it to daemon mode, and then fetch the segments sequentially inside it

mredolatti (changes requested):

splitio/sync/split.py
chillaq: The logic used to be in sync/synchronizer.py, Matti asked to move it to sync/split.py, that is why segment_sync object is needed. I will put a check before using it to avoid errors.

splitio/sync/split.py
chillaq: Hi Martin, this logic is also used for on-going sync, so we need to only sync the new segments added to splits, and ignore existing segments in storage. That is why we need to check for the segment before we sync

splitio/sync/synchronizer.py
chillaq: This was similar to the old logic before Matti asked to move it to sync/split.py

splitio/sync/synchronizer.py
chillaq: We are adding the new segments to the workerpool so the syncing will happen in bg, any advantage adding new code to perform the bg synching?

tests/sync/test_synchronizer.py
chillaq: First suggestion is applied in the test_synchronize_splits function, as the segment name is picked up from the split definitions.
Will create the second test.

splitio/sync/synchronizer.py
chillaq: Sorry, will check for such macro.

splitio/sync/synchronizer.py
chillaq: Looks like we are good with using workerpool and will only wait for completion when called from sync_now(), otherwise the workerpool will sync the segment(s) in the background, and the sync function return True

splitio/sync/synchronizer.py
chillaq: We will move with this recommendation.

splitio/sync/split.py
```diff
@@ -97,11 +99,11 @@ def _attempt_split_sync(self, fetch_options, till=None):
         remaining_attempts = _ON_DEMAND_FETCH_BACKOFF_MAX_RETRIES
         while True:
             remaining_attempts -= 1
-            change_number = self._fetch_until(fetch_options, till)
+            change_number, segment_list = self._fetch_until(fetch_options, till)
```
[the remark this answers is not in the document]
chillaq: We do accumulate in this line within the loop, the segments are added to the set.
 segment_list.update(set(parsed.get_segment_names()))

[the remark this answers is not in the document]
mredolatti: but not here.

There's a `while True`, which ends when change_number is greater than till.

if we call `_fetch_until` more than once, we'll lose the segment list returned by previous iterations and only return the latest. `segment_list` is overriden on each iteration

mredolatti (changes requested):

splitio/sync/synchronizer.py
```diff
@@ -233,7 +233,17 @@ def synchronize_splits(self, till):
...
+            for segment in self._split_synchronizers.split_sync.synchronize_splits(till):
+                    if not self._split_synchronizers.segment_sync.segment_exist_in_storage(segment):
+                        new_segments.append(segment)
+            if sync_segments and len(new_segments) != 0:
+                success = self._split_synchronizers.segment_sync.synchronize_segments(new_segments, True)
+                if not success:
+                    _LOGGER.error('Failed to schedule sync one or all segment(s) below.')
+                    _LOGGER.error(','.join(new_segments))
+                else:
+                    _LOGGER.debug('Segment sync scheduled.')
```
mredolatti, line 246: nit: we could print the list of scheduled segments here as well

mredolatti (approved):

PR #265, opened 2022-07-29
BloomFilter implementation

chillaq:
# Python SDK

## Extra Notes


setup.py
```diff
@@ -8,7 +8,7 @@
     'flake8',
     'pytest==7.0.1',
     'pytest-mock>=3.5.1',
-    'coverage==6.2',
+    'coverage==6.1',
```
sanzmauro, line 11: why do you downgrade this dependency?

splitio/engine/filters/base_filter.py
```diff
@@ -0,0 +1,28 @@
+import abc
+
+class BaseFilter(object, metaclass=abc.ABCMeta):
```
sanzmauro, line 3: rename to `Filter` please

splitio/engine/filters/bloom_filter.py
```diff
@@ -0,0 +1,53 @@
+from splitio.engine.filters.base_filter import BaseFilter
+from splitio import util
+from bloom_filter2 import BloomFilter
+
+class ImpressionsBloomFilter(BaseFilter):
```
sanzmauro, line 5: rename to `BloomFilter` please

splitio/engine/filters/base_filter.py
```diff
@@ -0,0 +1,28 @@
+import abc
```
sanzmauro, line 1: you are creating `__init__.py` file and is empty, is it correct? I think that you can define the interface in `__init__.py` and avoid to create this one.

sanzmauro (changes requested):

setup.py
chillaq: The tests were failing, I will change the version back to 6.2

splitio/engine/filters/base_filter.py
chillaq: This is how interfaces named in the SDK code

splitio/engine/filters/bloom_filter.py
chillaq: BlooomFilter is the 3rd party class name used, did not want to have collisions

splitio/engine/filters/base_filter.py
chillaq: will do

PR #267, opened 2022-08-09
Impressions sender

chillaq:
# Python SDK

## Extra Notes

### DIFF
--- a/splitio/api/telemetry.py
+++ b/splitio/api/telemetry.py
@@ -1,0 +1,50 @@
+"""Impressions API module."""
+
+import logging
+
+from splitio.api import APIException
+from splitio.api.client import HttpClientException
+from splitio.api.commons import headers_from_metadata
+
+_LOGGER = logging.getLogger(__name__)
+
+
+class TelemetryAPI(object):  # pylint: disable=too-few-public-methods
+    """Class that uses an httpClient to communicate with the Telemetry API."""
+
+    def __init__(self, client, apikey, sdk_metadata):
+        """
+        Class constructor.
+
+        :param client: HTTP Client responsble for issuing calls to the backend.
+        :type client: HttpClient
+        :param apikey: User apikey token.
+        :type apikey: string
+        """
+        self._client = client
+        self._apikey = apikey
+        self._metadata = headers_from_metadata(sdk_metadata)
+
+    def record_unique_keys(self, uniques):
+        """
+        Send unique keys to the backend.
+
+        :param uniques: Unique Keys
+        :type json
+        """
+        try:
+            response = self._client.post(
+                'keys',
+                '/ss',
+                self._apikey,
+                body=uniques,
+                extra_headers=self._metadata
+            )
+            if not 200 <= response.status_code < 300:
+                raise APIException(response.body, response.status_code)
+        except HttpClientException as exc:
+            _LOGGER.error(
+                'Error posting unique keys because an exception was raised by the HTTPClient'
+            )
+            _LOGGER.debug('Error: ', exc_info=True)
+            raise APIException('Unique keys not flushed properly.') from exc
--- a/splitio/engine/sender_adapters/in_memory_sender_adapter.py
+++ b/splitio/engine/sender_adapters/in_memory_sender_adapter.py
@@ -1,0 +1,41 @@
+import json
+
+from splitio.engine.sender_adapters import ImpressionsSenderAdapter
+
+class InMemorySenderAdapter(ImpressionsSenderAdapter):
+    """In Memory Impressions Sender Adapter class."""
+
+    def __init__(self, telemtry_http_client):
+        """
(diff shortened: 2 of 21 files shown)


gthea (commented):
Should this point to the `development` branch?

splitio/engine/unique_keys_tracker.py
```diff
@@ -0,0 +1,95 @@
...
+
+        :param key: key to be added to MTK list
+        :type key: int
+        :param feature_name: split name associated with the key
+        :type feature_name: str
+
+        :return: True if successful
+        :rtype: boolean
+        """
+        if self._filter.contains(feature_name+key):
```
mredolatti, line 48: this should be done within the lock context

splitio/engine/unique_keys_tracker.py
```diff
@@ -0,0 +1,95 @@
...
+        return True
+
+    def _get_dict_size(self):
+        """
+        Return the size of unique keys dictionary (number of keys in all features)
+
+        :return: dictionary set() items count
+        :rtype: int
+        """
+        total_size = 0
```
mredolatti, line 70: perhaps we should consider tracking the size on our own. it can get expensive to call get_treatment() in a loop if the amount of features is large

splitio/engine/sender_adapters/in_memory_sender_adapter.py
```diff
@@ -0,0 +1,41 @@
...
+
+        :return: unique keys JSON
+        :rtype: json
+        """
+        formatted_uniques = json.load('{keys: []}')
+        if len(uniques) == 0:
+            return formatted_uniques
+        for key in uniques:
+            formatted_uniques['keys'].append('{"f":"' + key +'", "ks:['+ json.dump(uniques[key])+']}')
+        return formatted_uniques
```
mredolatti, line 41:
```
return {
  'keys':  [{'f': feature, 'ks': list(keys) for feature, keys in uniques.items()}]
}
```

splitio/sync/unique_keys.py
```diff
@@ -0,0 +1,90 @@
...
+        """
+        cache_size = self._uniqe_keys_tracker._get_dict_size()
+        if cache_size <= self._max_bulk_size:
+            self._uniqe_keys_tracker._impressions_sender_adapter.record_unique_keys(self._uniqe_keys_tracker._cache)
+        else:
+            for bulk in self._split_cache_to_bulks():
+                self._uniqe_keys_tracker._impressions_sender_adapter.record_unique_keys(bulk)
+
+        with self._lock:
+            self._uniqe_keys_tracker._cache = {}
```
mredolatti, line 35: we should not access the cache directly here. If we want to fetch & remove all items at once, then we should add a method `fetch_and_clear`, `pop_all` (or similar) to the `UniqueKeysTracker` interface and provide a suitable implementation that handles locking properly. Here we're using an unrelated lock, which does not prevent race conditions with other mthods in the unique key tracker.

splitio/sync/unique_keys.py
```diff
@@ -0,0 +1,90 @@
...
+        self._lock = threading.RLock()
+
+    def SendAll(self):
+        """
+        Flush the unique keys dictionary to split back end.
+        Limit each post to the max_bulk_size value.
+
+        """
+        cache_size = self._uniqe_keys_tracker._get_dict_size()
+        if cache_size <= self._max_bulk_size:
```
mredolatti, line 28: is this defined?

mredolatti (changes requested):

splitio/api/telemetry.py
```diff
@@ -0,0 +1,50 @@
...
+    def record_unique_keys(self, uniques):
+        """
+        Send unique keys to the backend.
+
+        :param uniques: Unique Keys
+        :type json
+        """
+        try:
+            response = self._client.post(
+                'keys',
```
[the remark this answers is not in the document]
chillaq: Yep, will push the fix

splitio/sync/unique_keys.py
chillaq: Got it, will do.

splitio/engine/unique_keys_tracker.py
chillaq: will do, I thought since it is only reading, no need for lock.

splitio/engine/sender_adapters/in_memory_sender_adapter.py
chillaq: did few fixes on this yesterday, will push it along with your suggestions to none-implementation PR

splitio/engine/unique_keys_tracker.py
chillaq: I can increment the size by 1 if the unique key is added to dict, we won't need to get the size every time, but need to do that in a lock.

PR #269, opened 2022-08-24
Feature/mtk tracking

chillaq:
# Python SDK

## What did you accomplish?
NONE implementation for in memory
## Extra Notes


splitio/engine/unique_keys_tracker.py
```diff
@@ -0,0 +1,102 @@
...
+            self._add_or_update(feature_name, key)
+            self._filter.add(feature_name+key)
+            self._current_cache_size = self._current_cache_size + 1
+
+        if self._current_cache_size > self._cache_size:
+            _LOGGER.info(
+                'Unique Keys queue is full, flushing the current queue now.'
+            )
+            if self._queue_full_hook is not None and callable(self._queue_full_hook):
+                self._queue_full_hook()
```
mredolatti, line 62: should you clear the current_cache_size here?

splitio/engine/unique_keys_tracker.py
```diff
@@ -0,0 +1,102 @@
...
+        if callable(hook):
+            self._queue_full_hook = hook
+
+    def filter_pop_all(self):
+        """
+        Delete the filter items
+
+        """
+        with self._lock:
+            self._filter.clear()
```
mredolatti, line 93: or maybe here?

splitio/engine/unique_keys_tracker.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def get_cache_info_and_pop_all(self):
+        with self._lock:
+            temp_cach = self._cache.copy()
```
mredolatti, line 97: i don't think you need to do a copy if you're updating the old reference to a new empty dict. unless we have other references to this (which i hope we don't :) )

splitio/sync/impression.py
```diff
@@ -79,11 +79,15 @@ def __init__(self, impressions_api, impressions_manager):
...
         """
         self._impressions_api = impressions_api
-        self._impressions_manager = impressions_manager
+        self._impressions_counter = imp_counter
 
     def synchronize_counters(self):
         """Send impressions from both the failed and new queues."""
-        to_send = self._impressions_manager.get_counts()
+
+        if not isinstance(self._impressions_counter, Counter):
```
mredolatti, line 87: ?
does this accept soemthing other than a counter?

splitio/engine/sender_adapters/in_memory_sender_adapter.py
```diff
@@ -0,0 +1,41 @@
...
+
+        :param uniques: unique keys disctionary
+        :type uniques: Dictionary {'feature1': set(), 'feature2': set(), .. }
+
+        :return: unique keys JSON
+        :rtype: json
+        """
+        if len(uniques) == 0:
+            return json.loads('{"keys": []}')
+
```
mredolatti, line 38: this can be omitted. the expression below handles this case properly

splitio/engine/strategies/strategy_none_mode.py
```diff
@@ -0,0 +1,35 @@
...
+        Unique keys tracking are updated.
+
+        :param impressions: List of impression objects with attributes
+        :type impressions: list[tuple[splitio.models.impression.Impression, dict]]
+
+        :returns: Empty list, no impressions to post
+        :rtype: list[]
+        """
+        self._counter.track([imp for imp, _ in impressions])
+        [self._unique_keys_tracker.track(i.matching_key, i.feature_name) for i, _ in impressions]
```
mredolatti, line 34: this comprehensio allocates & populates a list that's not used. i'd say use a regular loop here:
```
for i, _ in impressions:
  self._unique_keys_tracker.track(i.matching_key, i.feature_name)
```

mredolatti (changes requested):

splitio/engine/unique_keys_tracker.py
chillaq: The hook function (which is send_all) calls get_cache_info_and_pop_all, which clears the cache and post a copy of it.

splitio/engine/unique_keys_tracker.py
chillaq: The hook function (which is send_all) calls get_cache_info_and_pop_all, which clears the cache and post a copy of it.

splitio/engine/unique_keys_tracker.py
chillaq: will do

splitio/sync/impression.py
chillaq: only counter, it is a check if the instance is not none.

mredolatti (approved):
lets put this in staging!

PR #273, opened 2022-09-16
Telemetry inmemory

chillaq:
# Python SDK

## What did you accomplish?
Core telemetry classes for in-memory

## Extra Notes


splitio/engine/impressions.py
```diff
@@ -1,160 +1,19 @@
 """Split evaluator module."""
-import threading
```
mredolatti, line 2: i'd still advise to use a nested package approach tu structure all this. The splitio/engine package is not meant for impressions only. it'd be ideal to have a sub-package impressions inside it ant the rest of the impressions-related modules inside it as follows:
```
splitio/engine/impressions/__init__.py
splitio/engine/impressions/adapters.py
splitio/engine/impressions/manager.py
splitio/engine/impressions/strategies.py
splitio/engine/impressions/unique_keys_tracker.py
```

the __init__.py could even contain a method that does most of the configuration, to prevent cluttering the factory.py with that many if/else branches according to the impression mode

splitio/engine/unique_keys_tracker.py
```diff
@@ -0,0 +1,103 @@
...
+        :type feature_name: str
+
+        :return: True if successful
+        :rtype: boolean
+        """
+        with self._lock:
+            if self._filter.contains(feature_name+key):
+                return False
+
+        with self._lock:
```
mredolatti, line 52: you don't need to separate `with lock` statements

splitio/engine/unique_keys_tracker.py
```diff
@@ -0,0 +1,103 @@
...
+    def set_queue_full_hook(self, hook):
+        """
+        Set a hook to be called when the queue is full.
+
+        :param h: Hook to be called when the queue is full
+        """
+        if callable(hook):
+            self._queue_full_hook = hook
+
+    def filter_pop_all(self):
```
mredolatti, line 88: this is a bit misleading, either the method should be called `clear_filter` or it should return the elements it's deleting

splitio/engine/unique_keys_tracker.py
```diff
@@ -0,0 +1,103 @@
...
+        Add the feature_name+key to both bloom filter and dictionary.
+
+        :param feature_name: split name associated with the key
+        :type feature_name: str
+        :param key: key to be added to MTK list
+        :type key: int
+        """
+        if feature_name not in self._cache:
+            self._cache[feature_name] = set()
+        self._cache[feature_name].add(key)
```
mredolatti, line 77: lock

splitio/recorder/recorder.py
```diff
@@ -129,6 +129,8 @@ def record_treatment_stats(self, impressions, latency, operation):
                 if self._data_sampling < rnumber:
                     return
             impressions = self._impressions_manager.process_impressions(impressions)
+            if impressions == []:
```
mredolatti, line 132: `if not impressions:`

splitio/storage/inmemmory.py
```diff
@@ -419,3 +452,263 @@ def clear(self):
...
+            self._config['sE'] = config['streamingEnabled']
+            self._config['rR'] = self._get_refresh_rates(config)
+            self._config['uO'] = self._get_url_overrides(config)
+            self._config['iQ'] = config['impressionsQueueSize']
+            self._config['eQ'] = config['eventsQueueSize']
+            self._config['iM'] = self._get_impressions_mode(config['impressionsMode'])
+            self._config['iL'] = True if config['impressionListener'] is not None else False
+            self._config['hp'] = self._check_if_proxy_detected()
+            self._config['aF'] = config['activeFactoryCount']
+            self._config['rF'] = config['redundantFactoryCount']
```
mredolatti, line 490: is this the same confir that the user passes when isntantiating the factory? if so, i'd suggest not to modify it with extra keys, but rather pass any extra data as seprate arguments (or another dict with contet info)

splitio/storage/inmemmory.py
```diff
@@ -419,3 +452,263 @@ def clear(self):
...
+                           'hL': {'sp': [], 'se': [], 'ms': [], 'im': [], 'ic': [], 'ev': [], 'te': [], 'to': []}}
+        self._exceptions = {'mE': {'t': 0, 'ts': 0, 'tc': 0, 'tcs': 0, 'tr': 0}}
+        self._records = {'IS': {'sp': 0, 'se': 0, 'ms': 0, 'im': 0, 'ic': 0, 'ev': 0, 'te': 0, 'to': 0},
+                         'sL': 0}
+        self._http_errors = {'sp': {}, 'se': {}, 'ms': {}, 'im': {}, 'ic': {}, 'ev': {}, 'te': {}, 'to': {}}
+        self._streaming_events = []
+        self._tags = []
+        self._integrations = {}
+        self._config = {'bT':0, 'nR':0, 'uC': 0}
+        self._map_latencies = {'Treatment': 't', 'Treatments': 'ts', 'TreatmentWithConfig': 'tc', 'TreatmentsWithConfig': 'tcs', 'Track': 'tr'}
```
mredolatti, line 473: if we're going to use dicts like this, we should probably have some methods like:

`_reset_counters()`, `_reset_latencies()`, etc.

so that we can call them upon construction and everytime the data gets popped. having the code repetaed it's very easy to miss a key if we do an update of htis logic

splitio/sync/manager.py
```diff
@@ -127,3 +128,44 @@ def _streaming_feedback_handler(self):
...
+        :type unique_keys_task: splitio.tasks.unique_keys_sync.UniqueKeysSyncTask
+
+        :param clear_filter_task: clear filter task instance
+        :type clear_filter_task: splitio.tasks.clear_filter_task.ClearFilterSynchronizer
+
+        """
+        self._ready_flag = True
+        self._synchronizer = synchronizer
+
+    def recreate(self):
```
mredolatti, line 149: ?

splitio/sync/synchronizer.py
```diff
@@ -201,6 +223,17 @@ def __init__(self, split_synchronizers, split_tasks):
...
         self._split_tasks = split_tasks
+        self._periodic_data_recording_tasks = [
+            self._split_tasks.impressions_task,
+            self._split_tasks.events_task
+        ]
+        if self._split_tasks.impressions_count_task:
+            self._periodic_data_recording_tasks.append(self._split_tasks.impressions_count_task)
+        if self._split_tasks.unique_keys_task is not None:
+            self._periodic_data_recording_tasks.append(self._split_tasks.unique_keys_task)
+        if self._split_tasks.clear_filter_task is not None:
```
mredolatti, line 234: you can drop the `is not None`

splitio/sync/unique_keys.py
```diff
@@ -0,0 +1,83 @@
...
+        :type uniqe_keys_tracker: splitio.engine.uniqur_key_tracker.UniqueKeysTracker
+        """
+        self._unique_keys_tracker = unique_keys_tracker
+
+    def clear_all(self):
+        """
+        Clear the bloom filter cache
+
+        """
+        self._unique_keys_tracker.filter_pop_all()
```
mredolatti, line 83: do we really need this class? can't we just have the AsyncTask periodically call `unique_keys_tracker.filter_pop_all()`?

mredolatti (changes requested):

splitio/client/factory.py
```diff
@@ -393,19 +437,69 @@ def _build_redis_factory(api_key, cfg):
...
+        imp_manager,
         storages['events'],
         storages['impressions'],
         data_sampling,
     )
+
+    manager = RedisManager(synchronizer)
+    initialization_thread = threading.Thread(target=manager.start, name="SDKInitializer")
+    initialization_thread.setDaemon(True)
+    initialization_thread.start()
```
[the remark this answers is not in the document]
chillaq: This is similar to the build factory in memory, the initialization_thread is initialized within the factory class, do you want to move both methods to their manager class?

splitio/engine/adapters.py
```diff
@@ -0,0 +1,146 @@
...
+        """
+        Build an impression bulk formatted as the API expects it.
+
+        :param counters: List of impression counters per feature.
+        :type counters: list[splitio.engine.impressions.Counter.CountPerFeature]
+
+        :return: dict with list of impression count dtos
+        :rtype: dict
+        """
+        return [json.dumps({
```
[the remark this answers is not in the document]
chillaq: This is identical code for impression counters to in-memory adapter, I am using it for Redis here.
It is returning only one item and its value is the counter array.
See https://github.com/splitio/python-client/blob/telemetry-inmemory/splitio/api/impressions.py#L77

splitio/sync/manager.py
chillaq: ._ready_flag is called from SplitFactory class; resume method (see https://github.com/splitio/python-client/blob/telemetry-inmemory/splitio/client/factory.py#L267)
I was not sure if customers can implement the fork workaround with redis mode.

splitio/sync/unique_keys.py
chillaq: The class is part of the specs to setup syncing for clearing the bloom filter, its the same design for all syncing tasks.

splitio/client/factory.py
```diff
@@ -453,9 +547,13 @@ def _build_localhost_factory(cfg):
 def get_factory(api_key, **kwargs):
     """Build and return the appropriate factory."""
     try:
+        active_factory_count = 0
+        redundant_factory_count = 0
         _INSTANTIATED_FACTORIES_LOCK.acquire()
         if _INSTANTIATED_FACTORIES:
             if api_key in _INSTANTIATED_FACTORIES:
+                redundant_factory_count = redundant_factory_count + 1
+                active_factory_count = active_factory_count + 1
```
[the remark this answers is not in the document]
chillaq: Correct, will move it out

splitio/client/factory.py
```diff
@@ -453,9 +547,13 @@ def _build_localhost_factory(cfg):
 def get_factory(api_key, **kwargs):
     """Build and return the appropriate factory."""
     try:
+        active_factory_count = 0
+        redundant_factory_count = 0
```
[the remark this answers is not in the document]
chillaq: That is what the code is doing, these variables are updated inside the lock block every time a new factory instance is created

splitio/storage/inmemmory.py
chillaq: will do

splitio/client/factory.py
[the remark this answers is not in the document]
mredolatti: those variables are scoped to the get_factory function. 

every time the function runs, the variables are set to 0. 
if this apikey has already been used, on line 555 & 556 you increment them both (now they're 1).
otherwise you only increment the active_Factory_count (now 1).

if you previously had 220 factories for the same apikey, the config will report 1 for active, and 1 for redundant.

active factories should be fetched from `len(_INSTANTIATED_FACTORIES)`
redundant factories for this apikey should be fetched from `_INSTANTIATED_FACTORIES[api_key]`

splitio/sync/manager.py
mredolatti: sorry, i meant to ask what is this `recreate` method used for?

splitio/sync/unique_keys.py
mredolatti: Let's chat offline on this one

splitio/sync/manager.py
chillaq: Same as above, the method recreate is called  with the fork workaround implementation.

splitio/client/factory.py
```diff
@@ -283,15 +292,22 @@ def _wrap_impression_listener(listener, metadata):
...
     """Build and return a split factory tailored to the supplied config."""
     if not input_validator.validate_factory_instantiation(api_key):
         return None
 
+    cfg['sdk_url'] = sdk_url if sdk_url is not None else None
+    cfg['events_url'] = events_url if events_url is not None else None
+    cfg['auth_url'] = auth_api_base_url if auth_api_base_url is not None else None
+    cfg['streaming_url'] = streaming_api_base_url if streaming_api_base_url is not None else None
+    cfg['telemetry_api_url'] = telemetry_api_base_url if telemetry_api_base_url is not None else None
+
```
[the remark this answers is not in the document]
chillaq: Done

splitio/client/factory.py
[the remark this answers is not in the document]
chillaq: Fixed

splitio/models/telemetry.py
```diff
@@ -25,3 +99,979 @@ def get_latency_bucket_index(micros):
...
+        """
+        with self._lock:
+            return {
+                BLOCK_UNTIL_READY_TIMEOUT:  self._block_until_ready_timeout,
+                NOT_READY: self._not_ready,
+                TIME_UNTIL_READY: self._time_until_ready,
+                OPERATION_MODE: self._operation_mode,
+                STORAGE_TYPE: self._storage_type,
+                STREAMING_ENABLED: self._streaming_enabled,
+                REFRESH_RATE: {'sp': self._refresh_rate.splits,
```
EmilianoSanchez, line 960: Just a comment that can help to remove some code: 
`TelemetryConfig` can have a reference to the SDK `config` object, and call `self._config[SPLITS_REFRESH_RATE]` instead of `self._refresh_rate.splits`, and the same for other refresh rates, URL overrides, _streaming_enabled, .... This way `_get_refresh_rates` method and `RefreshRates` class can be removed (idem URL related code).

EmilianoSanchez (changes requested):

splitio/models/telemetry.py
```diff
@@ -25,3 +99,979 @@ def get_latency_bucket_index(micros):
...
+        :type latency: int
+        """
+        with self._lock:
+            if method == TREATMENT:
+                if len(self._treatment) < MAX_LATENCY_BUCKET_COUNT:
+                    self._treatment.append(latency)
+            elif method == TREATMENTS:
+                if len(self._treatments) < MAX_LATENCY_BUCKET_COUNT:
+                    self._treatments.append(latency)
+            elif method == TREATMENT_WITH_CONFIG:
```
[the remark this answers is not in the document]
chillaq: That was the original code, Mati wanted to use methods instead of dict

splitio/models/telemetry.py
```diff
@@ -25,3 +99,979 @@ def get_latency_bucket_index(micros):
...
+        """
+        Append to the resource value
+
+        :param resource: passed resource name
+        :type resource: str
+        :param value: value to be appended
+        :type value: int
+        """
+        with self._lock:
+            if resource == IMPRESSIONS_QUEUED:
```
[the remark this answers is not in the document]
chillaq: That was the original code, Mati wanted to use methods instead of dict

splitio/models/telemetry.py
chillaq: That was the original code, Mati wanted to use methods instead of dict

mmelograno (changes requested):

splitio/models/telemetry.py
```diff
@@ -25,3 +99,969 @@ def get_latency_bucket_index(micros):
...
+
+    def append_token_refreshes(self):
+        """
+        Increament the token refreshes resource by one.
+
+        """
+        with self._lock:
+            self._token_refreshes = self._token_refreshes + 1
+
+    def set_value(self, resource, value):
```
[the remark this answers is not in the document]
chillaq: Will remove set method

splitio/models/telemetry.py
```diff
@@ -25,3 +99,969 @@ def get_latency_bucket_index(micros):
...
+        """
+        Add Latency method
+
+        :param resource: passed resource name
+        :type resource: str
+        :param latency: amount of latency
+        :type latency: int
+        """
+        with self._lock:
+            if resource == SPLIT:
```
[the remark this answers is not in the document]
chillaq: will rename latency to sync_time

splitio/models/telemetry.py
```diff
@@ -25,3 +99,969 @@ def get_latency_bucket_index(micros):
...
+            self._impressions_queued = 0
+            self._impressions_deduped = 0
+            self._impressions_dropped = 0
+            self._events_queued = 0
+            self._events_dropped = 0
+            self._auth_rejections = 0
+            self._token_refreshes = 0
+            self._session_length = 0
+
+    def append_value(self, resource, value):
```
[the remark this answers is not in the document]
chillaq: will rename append_ to record_

splitio/models/telemetry.py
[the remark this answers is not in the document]
chillaq: will remove set_ method

splitio/models/telemetry.py
```diff
@@ -25,3 +99,969 @@ def get_latency_bucket_index(micros):
...
+            elif resource == IMPRESSIONS_DROPPED:
+                return self._impressions_dropped
+            elif resource == EVENTS_QUEUED:
+                return self._events_queued
+            elif resource == EVENTS_DROPPED:
+                return self._events_dropped
+            else:
+                return 0
+
+    def get_session_length(self):
```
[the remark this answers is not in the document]
chillaq: This method has only get_

splitio/models/telemetry.py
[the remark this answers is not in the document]
chillaq: .. And split impressions and events to 2 methods

splitio/models/telemetry.py
```diff
@@ -25,3 +99,969 @@ def get_latency_bucket_index(micros):
...
+
+        :return: token refreshes value
+        :rtype: int
+        """
+        with self._lock:
+            token_refreshes = self._token_refreshes
+            self._token_refreshes = 0
+            return token_refreshes
+
+class StreamingEvent(object):
```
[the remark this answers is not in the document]
chillaq: will change param to tuple and remove lock

splitio/models/telemetry.py
```diff
@@ -25,3 +99,969 @@ def get_latency_bucket_index(micros):
...
+        """
+        Get telemetry refresh rate
+
+        :return: telemetry refresh rate
+        :rtype: int
+        """
+        with self._lock:
+            return self._telemetry
+
+class URLOverrides(object):
```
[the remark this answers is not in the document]
chillaq: will change to dict

splitio/models/telemetry.py
```diff
@@ -25,3 +99,969 @@ def get_latency_bucket_index(micros):
...
+        :return: streaming events dict
+        :rtype: dict
+        """
+
+        with self._lock:
+            streaming_events = self._streaming_events
+            self._streaming_events = []
+            return {STREAMING_EVENTS: [{'e': streaming_event.type, 'd': streaming_event.data,
+                                         't': streaming_event.time} for streaming_event in streaming_events]}
+class RefreshRates(object):
```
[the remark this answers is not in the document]
chillaq: will change to dict

splitio/models/telemetry.py
```diff
@@ -175,45 +171,39 @@ def __init__(self):
...
         """Reset variables"""
         with self._lock:
-            self._split = []
-            self._segment = []
-            self._impression = []
-            self._impression_count = []
-            self._event =[]
-            self._telemetry = []
-            self._token = []
+            self._split = [0] * 23
```
EmilianoSanchez, line 174: `[0] * 23` -> `[0] * MAX_LATENCY_BUCKET_COUNT` ?

EmilianoSanchez (approved):
LGTM

splitio/models/telemetry.py
chillaq: correct, will replace all 23

PR #279, opened 2022-10-19
Telemetry implementation for in-memory

chillaq:
# Python SDK

## Extra Notes


splitio/api/auth.py
```diff
@@ -42,7 +42,8 @@ def authenticate(self):
                 'auth',
                 '/v2/auth',
                 self._apikey,
-                extra_headers=self._metadata
+                extra_headers=self._metadata,
+                metric_name='token'
             )
             if 200 <= response.status_code < 300:
```
mmelograno, line 48: you can track auth_rejection on the else statement instead of passing trough metric_name `token`

mmelograno: and keep in mind that authrejections is only for 401 err

splitio/api/client.py
```diff
@@ -106,11 +107,22 @@ def get(self, server, path, apikey, query=None, extra_headers=None):  # pylint:
...
+            response = HttpResponse(response.status_code, response.text)
+            self._telemetry_runtime_producer.record_sync_latency(metric_name, elapsed)
+            if not 200 <= response.status_code < 300:
+                self._telemetry_runtime_producer.record_sync_error(metric_name, response.status_code)
+                if metric_name == 'token':
+                    self._telemetry_runtime_producer.record_auth_rejections()
+            else:
+                self._telemetry_runtime_producer.record_suceessful_sync(metric_name, round(1000 * elapsed))
+                if metric_name == 'token':
+                    self._telemetry_runtime_producer.record_token_refreshes()
```
mmelograno, line 120: token refresh should be tracked here https://github.com/splitio/python-client/blob/master/splitio/push/manager.py#L148

mmelograno: also succesful sync for auth

mmelograno (changes requested):
Latencies, Token Refresh, Auth Rejections and Successful Sync can be tracked in the outer component such as manager, synchronizers instead of the innest class as api

splitio/client/client.py
```diff
@@ -116,12 +117,13 @@ def _make_evaluation(self, key, feature, attributes, method_name, metric_name):
                 bucketing_key,
                 utctime_ms(),
             )
-
             self._record_stats([(impression, attributes)], start, metric_name)
```
mmelograno, line 120: I would encourage to use the record_stats method for tracking latency times on method. you have there you long it takes

splitio/client/factory.py
```diff
@@ -189,7 +201,15 @@ def block_until_ready(self, timeout=None):
             ready = self._sdk_ready_flag.wait(timeout)
 
             if not ready:
+                self._telemetry_init_producer.record_bur_time_out()
```
mmelograno, line 204: this should be record non ready usage, the bur timeouts is when the exception of timeout is raised

mmelograno (changes requested):

splitio/client/factory.py
chillaq: The exception line is below it though, so it is raising the timeout exception.

splitio/client/client.py
chillaq: I don't have method name though in record_stats, I would need to pass it

splitio/api/auth.py
chillaq: fixed

splitio/api/telemetry.py
```diff
@@ -33,18 +33,19 @@ def record_unique_keys(self, uniques):
...
+                metric_name='telemetry'
             )
             if not 200 <= response.status_code < 300:
                 raise APIException(response.body, response.status_code)
         except HttpClientException as exc:
             _LOGGER.info(
                 'Error posting unique keys because an exception was raised by the HTTPClient'
             )
-            _LOGGER.debug('Error: ', exc_info=True)
+            _LOGGER.info('Error: ', exc_info=True)
```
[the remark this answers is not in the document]
chillaq: fixed

splitio/api/auth.py
```diff
@@ -37,17 +39,21 @@ def authenticate(self):
...
                 self._apikey,
-                extra_headers=self._metadata
+                extra_headers=self._metadata,
             )
+            record_telemetry(response.status_code, int(round(time.time() * 1000)) - start, TOKEN, self._telemetry_runtime_producer)
             if 200 <= response.status_code < 300:
                 payload = json.loads(response.body)
                 return from_raw(payload)
             else:
+                if response.status_code == 401:
```
EmilianoSanchez, line 55: What about `if (response.status_code >= 400 && response.status_code < 500)` to cover all 4XX errors? (we are getting 401 for invalid api key, but 400 if using a client-side API Key on server-side)

splitio/api/commons.py
```diff
@@ -32,6 +32,27 @@ def headers_from_metadata(sdk_metadata, client_key=None):
...
+
+    :param metric_name: metric name for telemetry
+    :type metric_name: str
+
+    :param telemetry_runtime_producer: telemetry recording instance
+    :type telemetry_runtime_producer: splitio.engine.telemetry.TelemetryRuntimeProducer
+    """
+    telemetry_runtime_producer.record_sync_latency(metric_name, elapsed)
+    if 200 <= status_code < 300:
+        telemetry_runtime_producer.record_suceessful_sync(metric_name, int(round(time.time() * 1000)))
```
EmilianoSanchez, line 53: `record_suceessful_sync` -> `record_successful_sync`

splitio/api/impressions.py
```diff
@@ -91,14 +94,16 @@ def flush_impressions(self, impressions):
         :type impressions: list
         """
         bulk = self._build_bulk(impressions)
+        start = int(round(time.time() * 1000))
```
EmilianoSanchez, line 97: The expression `int(round(time.time() * 1000))` is being used in multiple places. 
If you think it might make your code simpler, consider defining some utility function.

Example in JS: https://github.com/splitio/javascript-commons/blob/development/src/utils/timeTracker/timer.ts

splitio/push/status_tracker.py
```diff
@@ -145,16 +151,20 @@ def _update_status(self):
         if self._last_status_propagated == Status.PUSH_SUBSYSTEM_UP:
             if not self._occupancy_ok() \
                     or self._last_control_message == ControlType.STREAMING_PAUSED:
+                self._telemetry_runtime_producer.record_streaming_event(('STREAMING_STATUS', 'PAUSED', self._timestamps))
```
EmilianoSanchez, line 154: Not so important, but consider defining string literals as constants to reuse across the code.

splitio/api/auth.py
chillaq: will do.

splitio/api/impressions.py
chillaq: will do.

splitio/push/status_tracker.py
chillaq: will do

splitio/client/client.py
```diff
@@ -393,7 +407,13 @@ def track(self, key, traffic_type, event_type, value=None, properties=None):
...
             properties=properties,
         )
-        return self._recorder.record_track_stats([EventWrapper(
+
+        return_flag = self._recorder.record_track_stats([EventWrapper(
             event=event,
             size=size,
         )])
+        self._telemetry_evaluation_producer.record_latency(TRACK, get_current_epoch_time() - start)
+        if not return_flag:
```
mmelograno, line 416: this is not required, keep in mind that track can return false and that won't be an exception. we will only return exception only in captured exception calls

splitio/sync/synchronizer.py
```diff
@@ -350,6 +364,7 @@ def stop_periodic_data_recording(self, blocking):
                 stop_event = threading.Event()
                 task.stop(stop_event)
                 events.append(stop_event)
+                time.sleep(0.4)
```
mmelograno, line 367: this should not be here

mmelograno (changes requested):

splitio/sync/synchronizer.py
chillaq: will add the fix proposed by Tincho

splitio/api/commons.py
chillaq: fixed typo

splitio/client/client.py
chillaq: Will add a try exception block.

splitio/models/telemetry.py
```diff
@@ -68,6 +67,25 @@
 TREATMENT_WITH_CONFIG = 'treatmentWithConfig'
 TREATMENTS_WITH_CONFIG = 'treatmentsWithConfig'
 TRACK = 'track'
+CONNECTION_ESTABLISHED = 'CONNECTION_ESTABLISHED'
```
mredolatti, line 70: given the large amounts of constants here, we could consider using enums to keep them scoped and more legible.
For example:

```
from enum import Enum

class ApiOperations(Enum):
  SPLIT_SYNC = 0
  SEGMENT_SYNC = 1
  ...


class ClientMethods(Enum):
  GET_TREATMENT = 0
  GET_TREATMENTS = 1
  ...
```

from other modules we would only import the enum class and use its values scoped:
```
from ...... import ClientMethods

record(ClientMethods.GET_TREATMENT, .....)
```

and when comparing within this module:

```
if method == ClientMethods.GET_TREATMENT:
   ...
elif method == ClientMethods.GET_TREATMENTS:
  ...
```

mmelograno (approved):

splitio/sync/synchronizer.py
```diff
@@ -225,7 +238,8 @@ def __init__(self, split_synchronizers, split_tasks):
         self._split_tasks = split_tasks
         self._periodic_data_recording_tasks = [
             self._split_tasks.impressions_task,
-            self._split_tasks.events_task
+            self._split_tasks.events_task,
+            self._split_tasks.telemetry_task
```
mredolatti, line 242: do we need this reference here? otherwise we can just remove it and then we don't have to check for it when iterating during shutdown.

mredolatti (changes requested):

splitio/client/factory.py
```diff
@@ -91,6 +94,9 @@ def __init__(  # pylint: disable=too-many-arguments
             recorder,
             sync_manager=None,
             sdk_ready_flag=None,
+            telemetry_producer=None,
```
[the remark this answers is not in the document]
chillaq: Will review all optional params and update accordingly

splitio/sync/synchronizer.py
chillaq: We need it when we start services, i can remove it from here and add a line in start services call

splitio/models/telemetry.py
chillaq: Will do

mredolatti (changes requested):

splitio/client/client.py
```diff
@@ -340,9 +347,12 @@ def _record_stats(self, impressions, start, operation):
         :param operation: operation performed.
         :type operation: str
         """
-        end = int(round(time.time() * 1000))
+        end = get_current_epoch_time()
         self._recorder.record_treatment_stats(impressions, get_latency_bucket_index(end - start),
                                               operation)
+        if not method_name == None:
```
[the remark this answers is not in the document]
chillaq: will change

splitio/client/client.py
```diff
@@ -393,7 +407,16 @@ def track(self, key, traffic_type, event_type, value=None, properties=None):
...
+                event=event,
+                size=size,
+            )])
+            self._telemetry_evaluation_producer.record_latency(MethodExceptionsAndLatencies.TRACK.value, get_current_epoch_time() - start)
+        except Exception:  # pylint: disable=broad-except
+            self._telemetry_evaluation_producer.record_exception(MethodExceptionsAndLatencies.TRACK.value)
+            _LOGGER.error('Error processing track event')
+            _LOGGER.debug('Error: ', exc_info=True)
+
+        return return_flag
```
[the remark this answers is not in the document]
chillaq: will do

splitio/models/telemetry.py
```diff
@@ -18,72 +18,122 @@
...
 
-class OperationMode(object):
-    """
-    Storage modes constants
-
-    """
-    MEMEORY = 'in-memory'
+class OperationMode(Enum):
+    """Storage modes constants"""
+    MEMEORY = 'inmemory'
```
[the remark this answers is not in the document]
chillaq: will fix

splitio/models/telemetry.py
```diff
@@ -570,7 +620,9 @@ def __init__(self, streaming_event):
         :param streaming_event: Streaming event tuple: ('type', 'data', 'time')
         :type streaming_event: dict
         """
-        self._type, self._data, self._time = streaming_event
+        self._type = streaming_event[0]
```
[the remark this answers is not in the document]
chillaq: Actually I need the only the value for the _type:
        self._type = streaming_event[0].value
        self._data = streaming_event[1]
        self._time = streaming_event[2]

splitio/storage/inmemmory.py
```diff
@@ -420,10 +428,13 @@ def put(self, events):
                     if self._size >= MAX_SIZE_BYTES:
                         self._queue_full_hook()
                         return False
-
                     self._events.put(event.event, False)
+                    events_stored = events_stored + 1
```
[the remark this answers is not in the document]
chillaq: fixed all of them.

splitio/storage/inmemmory.py
```diff
@@ -511,13 +526,13 @@ def record_event_stats(self, data_type, count):
...
 
-    def record_suceessful_sync(self, resource, time):
+    def record_successful_sync(self, resource, time):
         """Record successful sync."""
         self._last_synchronization.add_latency(resource, time)
 
     def record_sync_error(self, resource, status):
         """Record sync http error."""
-        self._http_sync_errors.add_error(resource, status)
+        self._http_sync_errors.add_error(resource, str(status))
```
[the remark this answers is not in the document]
chillaq: It can be passed as an integer, but the storage is string.

splitio/sync/manager.py
```diff
@@ -17,7 +19,7 @@ class Manager(object):  # pylint:disable=too-many-instance-attributes
 
     _CENTINEL_EVENT = object()
 
-    def __init__(self, ready_flag, synchronizer, auth_api, streaming_enabled, sdk_metadata, sse_url=None, client_key=None):  # pylint:disable=too-many-arguments
+    def __init__(self, ready_flag, synchronizer, auth_api, streaming_enabled, sdk_metadata, telemetry_runtime_producer=None, sse_url=None, client_key=None):  # pylint:disable=too-many-arguments
```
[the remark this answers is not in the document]
chillaq: will make it mandatory

splitio/engine/telemetry.py
```diff
@@ -152,21 +155,21 @@ def get_config_stats(self):
...
+            'eQ': config_stats['eQ'],
+            'iM': config_stats['iM'],
+            'iL': config_stats['iL'],
+            'hp': config_stats['hp'],
+            'aF': config_stats['aF'],
+            'rF': config_stats['rF'],
+            'bT': config_stats['bT'],
+            'nR': config_stats['nR'],
+            'tR': config_stats['tR']}
         )
```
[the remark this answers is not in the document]
chillaq: constructing the telemetry config stats dict

splitio/api/commons.py
```diff
@@ -93,3 +114,12 @@ def build_fetch(change_number, fetch_options, metadata):
     if fetch_options.change_number is not None:
         query['till'] = fetch_options.change_number
     return query, extra_headers
+
+def get_current_epoch_time():
```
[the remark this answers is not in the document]
chillaq: will move it.

splitio/models/telemetry.py
```diff
@@ -18,72 +18,115 @@
...
+    IMPRESSION = 'impression'
+    IMPRESSION_COUNT = 'impressionCount'
+    EVENT = 'event'
+    TELEMETRY = 'telemetry'
+    TOKEN = 'token'
+
+class MethodExceptionsAndLatencies(Enum):
+    """Method exceptions and latencies constants"""
+    METHOD_LATENCIES = 'methodLatencies'
+    METHOD_EXCEPTIONS = 'methodExceptions'
```
mredolatti, line 79: i'd say these 2 should be separate constants, or at least not be mixed with the method names

splitio/push/status_tracker.py
```diff
@@ -73,11 +76,16 @@ def handle_occupancy(self, event):
...
+            _LOGGER.info('received an old occupancy message. ignoring.')
             return None
         self._timestamps.occupancy = event.timestamp
 
         self._publishers[event.channel] = event.publishers
+        if event.channel[-3:] == 'pri':
+            event_type = StreamingEventTypes.OCCUPANCY_PRI
+        else:
+            event_type = StreamingEventTypes.OCCUPANCY_SEC
+        self._telemetry_runtime_producer.record_streaming_event((event_type, len(self._publishers), event.timestamp))
```
mredolatti, line 88:
```
self._telemetry_runtime_producer.record_streaming_event((
    StreamingEventTypes.OCCUPANCY_PRI if event.channel[-3:] == 'pri' else StreamingEventTypes.OCCUPANCY_SEC,
    len(self._publishers),
    event.timestamp
))
```

splitio/storage/inmemmory.py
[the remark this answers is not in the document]
mredolatti: i think it'd be better to just use integers here, and then have the storage turn it into whatever it may need.

mredolatti (approved):

splitio/models/telemetry.py
chillaq: They both use the same exact method names, that is why I had them in the same class

splitio/push/status_tracker.py
chillaq: will do

PR #284, opened 2022-11-29
Fire timeout exception with invalid apikey

chillaq:
1- Added backoff to sync_all that fixes the timeout when invalid apikey is used
2- Removed api key validation

# Python SDK

## Extra Notes


splitio/sync/synchronizer.py
```diff
@@ -291,6 +299,8 @@ def sync_all(self):
             try:
                 if not self.synchronize_splits(None, False):
                     attempts -= 1
+                    how_long = self._backoff.get()
```
mmelograno, line 302: if we are moving to backoff, there won't be attempts any longer. you can remove the attempts

mmelograno (changes requested):

mredolatti (approved):

splitio/sync/synchronizer.py
chillaq: done

PR #305, opened 2023-02-03
added sha check and integration tests

chillaq:
# Python SDK

## Extra Notes


splitio/sync/segment.py
```diff
@@ -230,17 +231,20 @@ def synchronize_segment(self, segment_name, till=None):
         try:
             fetched = self._read_segment_from_json_file(segment_name)
             if not self.segment_exist_in_storage(segment_name):
+                    self._segment_sha[segment_name] = self._get_sha(json.dumps(fetched))
```
mmelograno, line 234: you are getting the `self._get_sha(json.dumps(fetched))` in several places over an over. Maybe you can get sha at the beginning

splitio/sync/split.py
```diff
@@ -351,24 +353,21 @@ def _synchronize_json(self):
...
-            for split in fetched:
-                parsed = splits.from_raw(split)
-                _LOGGER.debug("split %s is updated", parsed.name)
-                self._split_storage.put(parsed)
-
-                segment_list.update(set(parsed.get_segment_names()))
-
-            for split in to_delete:
-                self._split_storage.remove(split)
+        if self._get_sha(json.dumps(fetched)) != self._current_json_sha:
```
mmelograno, line 358: getting `self._get_sha(json.dumps(fetched)` on a variable will be better than calling twice

mmelograno (changes requested):

mmelograno (approved):

splitio/sync/segment.py
chillaq: done

PR #309, opened 2023-02-09
Allow sync for splits and segments when till=-1

chillaq:
# Python SDK

## What did you accomplish?
Allow sync for splits and segments when till=-1
Updated version
Set the default for telemetry sync to 1 hour and min to 1 min

## Extra Notes


nmayorsplit (approved):
LGTM

PR #311, opened 2023-02-14
Localhost json release

chillaq:
# Python SDK

## Extra Notes


splitio/client/config.py
```diff
@@ -123,4 +125,7 @@ def sanitize(apikey, config):
                                                     config.get('impressionsRefreshRate'))
     processed['impressionsMode'] = imp_mode
     processed['impressionsRefreshRate'] = imp_rate
+    if processed['metricsRefreshRate'] < 60:
+        processed['metricsRefreshRate'] = 60
```
mmelograno, line 129: 3600

splitio/client/factory.py
```diff
@@ -528,22 +527,37 @@ def _build_localhost_factory(cfg):
...
+                               LocalhostMode.JSON if cfg['splitFile'][-5:].lower() == '.json' else LocalhostMode.LEGACY),
+        LocalSegmentSynchronizer(cfg['segmentDirectory'], storages['splits'], storages['segments']),
+        None, None, None,
     )
 
-    tasks = SplitTasks(
-        SplitSynchronizationTask(
+    split_sync_task = None
+    segment_sync_task = None
+    if cfg['localhostRefreshEnabled']:
```
mmelograno, line 539: maybe you need to verified at this point if it's also JSON mode, I think this will be fired if customer enables with yaml

splitio/sync/segment.py
```diff
@@ -174,13 +175,153 @@ def synchronize_segments(self, segment_names = None, dont_wait = False):
...
+
+        :param parsed: segment dict
+        :type parsed: Dict
+
+        :return: sanitized segment structure dict
+        :rtype: Dict
+        """
+        if 'name' not in parsed or parsed['name'] is None:
+            _LOGGER.warning("Segment does not have [name] element, skipping")
+            return {}
```
mmelograno, line 311: wouldn't be better returning None instead of empty object?

splitio/sync/segment.py
```diff
@@ -174,13 +175,153 @@ def synchronize_segments(self, segment_names = None, dont_wait = False):
...
+        if parsed['name'].strip() == '':
+            _LOGGER.warning("Segment [name] element is blank, skipping")
+            return {}
```
mmelograno, line 314: wouldn't be better returning None instead of empty object?

splitio/sync/segment.py
```diff
@@ -174,13 +175,153 @@ def synchronize_segments(self, segment_names = None, dont_wait = False):
...
+
+        :param till: ChangeNumber received.
+        :type till: int
+
+        :return: True if no error occurs. False otherwise.
+        :rtype: bool
+        """
+        try:
+            fetched = self._read_segment_from_json_file(segment_name)
+            if fetched == {}:
```
mmelograno, line 257: wouldn't be better to check for None?

splitio/sync/split.py
```diff
@@ -19,7 +23,7 @@
 
 
 _ON_DEMAND_FETCH_BACKOFF_BASE = 10  # backoff base starting at 10 seconds
-_ON_DEMAND_FETCH_BACKOFF_MAX_WAIT = 60  # don't sleep for more than 1 minute
+_ON_DEMAND_FETCH_BACKOFF_MAX_WAIT = 30  # don't sleep for more than 1 minute
```
mmelograno, line 26: why we change this? comment should be 30 secs instead of 1 minute

splitio/sync/split.py
```diff
@@ -307,6 +322,27 @@ def _read_splits_from_yaml_file(cls, filename):
...
+        try:
+            if self._localhost_mode == LocalhostMode.JSON:
+                return self._synchronize_json()
+            else:
+                return self._synchronize_legacy()
+        except Exception as exc:
+            _LOGGER.error(str(exc))
+            raise APIException("Error fetching splits information") from exc
+
+#            _LOGGER.error("Error fetching splits information")
```
mmelograno, line 334: maybe remove this comments?

splitio/sync/split.py
```diff
@@ -318,3 +354,168 @@ def synchronize_splits(self, till=None):  # pylint:disable=unused-argument
...
+        :param split: split dict object
+        :type split: Dict
+
+        :return: sanitized split
+        :rtype: Dict
+        """
+        found_all_keys_matcher = False
+        if 'conditions' not in split or split['conditions'] is None:
+            split['conditions'] = []
+        for condition in split['conditions']:
```
mmelograno, line 484: you need to check if the last condition matches with ALL_KEYS not if it has one.

mmelograno (changes requested):

splitio/client/config.py
chillaq: will fix

splitio/sync/segment.py
chillaq: Possible, I am checking if the returned is {} then synchronize_segments() method will return False

splitio/sync/segment.py
chillaq: Possible, I am checking if the returned is {} then synchronize_segments() method will return False

splitio/sync/split.py
chillaq: I am using the same constant for both sync.split.py and sync.segment.py, I think it can be the same amount for both.

chillaq: will fix the comment.

splitio/client/factory.py
chillaq: will fix

splitio/client/client.py
```diff
@@ -115,8 +115,9 @@ def _make_evaluation(self, key, feature, attributes, method_name, metric_name):
             )
             self._record_stats([(impression, attributes)], start, metric_name, method_name)
             return result['treatment'], result['configurations']
-        except Exception:  # pylint: disable=broad-except
+        except Exception as e:  # pylint: disable=broad-except
             _LOGGER.error('Error getting treatment for feature')
+            _LOGGER.error(str(e))
```
mmelograno, line 120: do we need this?

splitio/sync/split.py
```diff
@@ -318,3 +354,172 @@ def synchronize_splits(self, till=None):  # pylint:disable=unused-argument
...
+        :rtype: Dict
+        """
+        found_all_keys_matcher = False
+        if 'conditions' not in split or split['conditions'] is None:
+            split['conditions'] = []
+        conditions_count = len(split['conditions'])
+        count = 0
+        for condition in split['conditions']:
+            count += 1
+            if count == conditions_count: # checking only last condition
```
mmelograno, line 488: I think that 
```
if len(split['conditions'] > 0:
   last = split['conditions'][-1]
   ...
```
would be better than iterate over all the elements O(n)

splitio/sync/synchronizer.py
```diff
@@ -505,40 +503,71 @@ def __init__(self, split_synchronizers, split_tasks):
...
-            raise APIException('Failed to sync splits') from exc
+        self._backoff.reset()
+        remaining_attempts = _ON_DEMAND_FETCH_BACKOFF_MAX_RETRIES
+        while remaining_attempts > 0:
+            remaining_attempts -= 1
+            try:
+                return self.synchronize_splits()
+            except APIException as exc:
+                _LOGGER.error('Failed syncing all')
+                _LOGGER.error(str(exc))
```
mmelograno, line 522: is this needed?

splitio/sync/synchronizer.py
```diff
@@ -505,40 +503,71 @@ def __init__(self, split_synchronizers, split_tasks):
...
-    def synchronize_splits(self, till):
+    def synchronize_splits(self):
         """Synchronize all splits."""
-        raise NotImplementedError()
+        try:
+            new_segments = []
+            for segment in self._split_synchronizers.split_sync.synchronize_splits():
+                    if not self._split_synchronizers.segment_sync.segment_exist_in_storage(segment):
+                        new_segments.append(segment)
+            if len(new_segments) != 0:
```
mmelograno, line 554: len(new_segments) > 0 ?

mmelograno (changes requested):

splitio/client/client.py
chillaq: I think so, I had to add this line to know why the treatment evaluation is failing, we are not showing it in the exception block

splitio/sync/split.py
chillaq: Yeah.. duh..

splitio/sync/synchronizer.py
chillaq: Yes, if there is any i/o or json error, it will show up with this line.

splitio/sync/synchronizer.py
chillaq: sure, will apply it.

splitio/sync/segment.py
chillaq: fixed

splitio/sync/split.py
chillaq: fixed

splitio/sync/segment.py
chillaq: fixed

mmelograno (approved):

PR #313, opened 2023-02-16
Added split storage class and test

chillaq:
# Python SDK

## Extra Notes

### DIFF
--- a/splitio/storage/pluggable.py
+++ b/splitio/storage/pluggable.py
@@ -10,12 +10,16 @@ _LOGGER = logging.getLogger(__name__)
 class PluggableSplitStorage(SplitStorage):
     """InMemory implementation of a split storage."""
 
-    def __init__(self, pluggable_adapter, prefix):
+    def __init__(self, pluggable_adapter, prefix=None):
         """Constructor."""
         self._pluggable_adapter = pluggable_adapter
-        self._prefix = prefix + ".split."
-        self._traffic_type_prefix = prefix + ".trafficType."
-        self._split_till_prefix = prefix + ".splits.till"
+        self._prefix = "split."
+        self._traffic_type_prefix = "trafficType."
+        self._split_till_prefix = "splits.till"
+        if prefix is not None:
+            self._prefix = prefix + "." + self._prefix
+            self._traffic_type_prefix = prefix + "." + self._traffic_type_prefix
+            self._split_till_prefix = prefix + "." + self._split_till_prefix
 
     def get(self, split_name):
         """
@@ -31,6 +35,10 @@ class PluggableSplitStorage(SplitStorage):
             return None
         return splits.from_raw(split)
 
+    def _get_key(self, key):
+        """Retrieve a key content."""
+        return self._pluggable_adapter.get(key)
+
     def fetch_many(self, split_names):
         """
         Retrieve splits.
@@ -41,7 +49,8 @@ class PluggableSplitStorage(SplitStorage):
         :return: A dict with split objects parsed from queue.
         :rtype: dict(split_name, splitio.models.splits.Split)
         """
-        return {split_name: self.get(split_name) for split_name in split_names}
+        prefix_added = [self._prefix + split for split in split_names]
+        return {split['name']: splits.from_raw(split) for split in self._pluggable_adapter.get_many(prefix_added)}
 
     def put_many(self, splits, change_number):
         """
@@ -70,8 +79,7 @@ class PluggableSplitStorage(SplitStorage):
             return False
 
         self._pluggable_adapter.delete(self._prefix + split_name)
-        self._pluggable_adapter.decrement(self._traffic_type_prefix + split.traffic_type_name, 1)
-        if self._pluggable_adapter.get(self._traffic_type_prefix + split.traffic_type_name) == 0:
+        if self._decrease_traffic_type_count(split.traffic_type_name) == 0:
             self._pluggable_adapter.delete(self._traffic_type_prefix + split.traffic_type_name)
         return True
 
@@ -101,7 +109,6 @@ class PluggableSplitStorage(SplitStorage):
         """
         return [split.name for split in self.get_all()]
 
-
     def get_all(self):
         """
         Return all the splits.
@@ -109,7 +116,7 @@ class PluggableSplitStorage(SplitStorage):
         :return: List of all the splits.
         :rtype: list
         """
-        return [splits.from_raw(split) for split in self._pluggable_adapter.get_keys_by_prefix(self._prefix)]
+        return [splits.from_raw(self._get_key(key)) for key in self._pluggable_adapter.get_keys_by_prefix(self._prefix)]
 
     def traffic_type_exists(self, traffic_type_name):
         """
@@ -141,27 +148,33 @@ class PluggableSplitStorage(SplitStorage):
             return
         split.local_kill(default_treatment, change_number)
         self._pluggable_adapter.set(self._prefix + split_name, split.to_json())
-        self.set_change_number(change_number)
 
-    def increase_traffic_type_count(self, traffic_type_name):
+    def _increase_traffic_type_count(self, traffic_type_name):
         """
         Increase by one the count for a specific traffic type name.
 
         :param traffic_type_name: Traffic type to increase the count.
         :type traffic_type_name: str
-        """
-        self._pluggable_adapter.increment(self._traffic_type_prefix + traffic_type_name, 1)
 
-    def decrease_traffic_type_count(self, traffic_type_name):
+        :return: existing count of traffic type
+        :rtype: int
+        """
+        return self._pluggable_adapter.increment(self._traffic_type_prefix + traffic_type_name, 1)
+
+    def _decrease_traffic_type_count(self, traffic_type_name):
         """
         Decrease by one the count for a specific traffic type name.
 
         :param traffic_type_name: Traffic type to decrease the count.
         :type traffic_type_name: str
+
+        :return: existing count of traffic type
+        :rtype: int
         """
-        self._pluggable_adapter.decrement(self._traffic_type_prefix + traffic_type_name, 1)
+        return_count = self._pluggable_adapter.decrement(self._traffic_type_prefix + traffic_type_name, 1)
         if self._pluggable_adapter.get(self._traffic_type_prefix + traffic_type_name) == 0:
             self._pluggable_adapter.delete(self._traffic_type_prefix + traffic_type_name)
+        return return_count
 
     def get_all_splits(self):
         """
@@ -191,5 +204,8 @@ class PluggableSplitStorage(SplitStorage):
         :param split: Split object.
         :type split: splitio.models.split.Split
         """
+        existing_split = self.get(split.name)
+        if existing_split is not None:
+            self._decrease_traffic_type_count(existing_split.traffic_type_name)
         self._pluggable_adapter.set(self._prefix + split.name, split.to_json())
-        self._pluggable_adapter.increment(self._traffic_type_prefix + split.traffic_type_name, 1)
+        self._increase_traffic_type_count(split.traffic_type_name)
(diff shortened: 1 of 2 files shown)


splitio/storage/pluggable.py
```diff
@@ -0,0 +1,195 @@
...
+
+
+    def get_all(self):
+        """
+        Return all the splits.
+
+        :return: List of all the splits.
+        :rtype: list
+        """
+        return [splits.from_raw(split) for split in self._pluggable_adapter.get_keys_by_prefix(self._prefix)]
```
EmilianoSanchez, line 112: Check here since `get_keys_by_prefix` is returning keys rather than split definitions.

splitio/storage/pluggable.py
```diff
@@ -0,0 +1,195 @@
...
+        :type change_number: int
+        """
+        split = self.get(split_name)
+        if not split:
+            return
+        if self.get_change_number() > change_number:
+            return
+        split.local_kill(default_treatment, change_number)
+        self._pluggable_adapter.set(self._prefix + split_name, split.to_json())
+        self.set_change_number(change_number)
```
EmilianoSanchez, line 144: Check if setting the change number here should be removed, to not mess up the flow of a SPLIT_KILL streaming notification: when a SPLIT_KILL notification arrives, the split must be killed locally immediately and also call `/splitChanges`, which in turn will update the change number.

splitio/storage/pluggable.py
```diff
@@ -0,0 +1,195 @@
...
+
+    def increase_traffic_type_count(self, traffic_type_name):
```
EmilianoSanchez, line 146: I think you can remove `increase_traffic_type_count` and `decrease_traffic_type_count` methods, since `put` and `remove` are already incrementing and decrementing the traffic type. Or you can use them in `put` and `remove` respectively.

splitio/storage/pluggable.py
```diff
@@ -0,0 +1,195 @@
...
+
+    def put(self, split):
+        """
+        Store a split.
+
+        :param split: Split object.
+        :type split: splitio.models.split.Split
+        """
+        self._pluggable_adapter.set(self._prefix + split.name, split.to_json())
+        self._pluggable_adapter.increment(self._traffic_type_prefix + split.traffic_type_name, 1)
```
EmilianoSanchez, line 195: You need to consider to decrement the traffic type of previous split in `_pluggable_adapter.get(self._prefix + split.name)`

splitio/storage/pluggable.py
EmilianoSanchez: As a reference, you can check what we do in JS https://github.com/splitio/javascript-commons/blob/development/src/storages/pluggable/SplitsCachePluggable.ts#L31

EmilianoSanchez (changes requested):

splitio/storage/pluggable.py
```diff
@@ -0,0 +1,195 @@
...
+
+_LOGGER = logging.getLogger(__name__)
+
+class PluggableSplitStorage(SplitStorage):
+    """InMemory implementation of a split storage."""
+
+    def __init__(self, pluggable_adapter, prefix):
+        """Constructor."""
+        self._pluggable_adapter = pluggable_adapter
+        self._prefix = prefix + ".split."
```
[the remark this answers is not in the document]
chillaq: correct, will fix it.

splitio/storage/pluggable.py
```diff
@@ -0,0 +1,195 @@
...
+        """
+        Retrieve splits.
+
+        :param split_names: Names of the features to fetch.
+        :type split_name: list(str)
+
+        :return: A dict with split objects parsed from queue.
+        :rtype: dict(split_name, splitio.models.splits.Split)
+        """
+        return {split_name: self.get(split_name) for split_name in split_names}
```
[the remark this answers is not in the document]
chillaq: I know the spec says "getMany", but Python SDK already use fetch_many for InMemory and Redis storages, and that method is called from other classes

splitio/storage/pluggable.py
```diff
@@ -0,0 +1,195 @@
...
+        :return: True if the split was found and removed. False otherwise.
+        :rtype: bool
+        """
+        split = self.get(split_name)
+        if not split:
+            _LOGGER.warning("Tried to delete nonexistant split %s. Skipping", split_name)
+            return False
+
+        self._pluggable_adapter.delete(self._prefix + split_name)
+        self._pluggable_adapter.decrement(self._traffic_type_prefix + split.traffic_type_name, 1)
```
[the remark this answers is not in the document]
chillaq: I don't see in the specs that decrement returns a value, can you update the spec?

splitio/storage/pluggable.py
chillaq: that is correct, will fix it.

splitio/storage/pluggable.py
chillaq: I don't see calling /splitChanges in Python InMemory class, looks like the sync class is only calling this method

splitio/storage/pluggable.py
chillaq: that is correct, I will reference them from put and remove, since they are in the spec

splitio/storage/pluggable.py
chillaq: do you mean if the existing split has different traffic type before it gets updated?

splitio/storage/pluggable.py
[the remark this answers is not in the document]
chillaq: will add get_many to the wrapper call

splitio/storage/pluggable.py
chillaq: will need to check if split exist and decrement the old tt and increment new tt

splitio/storage/pluggable.py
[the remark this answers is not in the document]
chillaq: I see it in the storage method list, will update the code

splitio/storage/pluggable.py
```diff
@@ -10,12 +10,16 @@
...
 
-    def __init__(self, pluggable_adapter, prefix):
+    def __init__(self, pluggable_adapter, prefix=None):
         """Constructor."""
         self._pluggable_adapter = pluggable_adapter
-        self._prefix = prefix + ".split."
-        self._traffic_type_prefix = prefix + ".trafficType."
-        self._split_till_prefix = prefix + ".splits.till"
+        self._prefix = "split."
+        self._traffic_type_prefix = "trafficType."
```
EmilianoSanchez, line 17: Shouldn't it be ?
```
self._traffic_type_prefix = self._prefix + "trafficType."
self._split_till_prefix = self._prefix + "splits.till"
```

splitio/storage/pluggable.py
```diff
@@ -31,6 +35,10 @@ def get(self, split_name):
             return None
         return splits.from_raw(split)
 
+    def _get_key(self, key):
```
EmilianoSanchez, line 38: Do we need this extra method in `PluggableSplitStorage`, when you can call `self._pluggable_adapter.get(key)` directly in line 119 ?

splitio/storage/pluggable.py
```diff
@@ -141,27 +148,33 @@ def kill_locally(self, split_name, default_treatment, change_number):
...
 
         :param traffic_type_name: Traffic type to decrease the count.
         :type traffic_type_name: str
+
+        :return: existing count of traffic type
+        :rtype: int
         """
-        self._pluggable_adapter.decrement(self._traffic_type_prefix + traffic_type_name, 1)
+        return_count = self._pluggable_adapter.decrement(self._traffic_type_prefix + traffic_type_name, 1)
         if self._pluggable_adapter.get(self._traffic_type_prefix + traffic_type_name) == 0:
```
EmilianoSanchez, line 175: `if self._pluggable_adapter.get(self._traffic_type_prefix + traffic_type_name) == 0:` --> 
`if return_count == 0:`

splitio/storage/pluggable.py
```diff
@@ -70,8 +79,7 @@ def remove(self, split_name):
             return False
 
         self._pluggable_adapter.delete(self._prefix + split_name)
-        self._pluggable_adapter.decrement(self._traffic_type_prefix + split.traffic_type_name, 1)
-        if self._pluggable_adapter.get(self._traffic_type_prefix + split.traffic_type_name) == 0:
+        if self._decrease_traffic_type_count(split.traffic_type_name) == 0:
             self._pluggable_adapter.delete(self._traffic_type_prefix + split.traffic_type_name)
```
EmilianoSanchez, line 83: No need to delete here, since `_decrease_traffic_type_count` is already deleting it if it gets to 0.

splitio/storage/pluggable.py
```diff
@@ -191,5 +204,8 @@ def put(self, split):
         :param split: Split object.
         :type split: splitio.models.split.Split
         """
+        existing_split = self.get(split.name)
+        if existing_split is not None:
+            self._decrease_traffic_type_count(existing_split.traffic_type_name)
         self._pluggable_adapter.set(self._prefix + split.name, split.to_json())
-        self._pluggable_adapter.increment(self._traffic_type_prefix + split.traffic_type_name, 1)
+        self._increase_traffic_type_count(split.traffic_type_name)
```
EmilianoSanchez, line 211: I would call `_increase_traffic_type_count` first, and then `_decrease_traffic_type_count`, to avoid unnecessary wrapper `delete` operations when calling `_decrease_traffic_type_count`

tests/storage/test_pluggable.py
```diff
@@ -15,6 +15,9 @@ def get(self, key):
             return None
         return self._keys[key]
 
+    def get_many(self, keys):
+        return [self.get(key) for key in keys]
+
     def set(self, key, value):
         self._keys[key] = value
```
EmilianoSanchez, line 23: Check `increment` and `decrement` methods, to return the updated value.
Example:
```
    def decrement(self, key, value):
        if key not in self._keys:
            self._keys[key] = 0
        self._keys[key]-= value
        return self._keys[key]
```

splitio/storage/pluggable.py
```diff
@@ -10,12 +10,16 @@
...
     """InMemory implementation of a split storage."""
 
-    def __init__(self, pluggable_adapter, prefix):
+    def __init__(self, pluggable_adapter, prefix=None):
         """Constructor."""
         self._pluggable_adapter = pluggable_adapter
-        self._prefix = prefix + ".split."
-        self._traffic_type_prefix = prefix + ".trafficType."
-        self._split_till_prefix = prefix + ".splits.till"
+        self._prefix = "split."
```
EmilianoSanchez, line 16: I now remember from JS, that the default prefix is `"SPLITIO."` (in uppercase), as you can check here https://github.com/splitio/python-client/blob/master/splitio/storage/redis.py#L21

EmilianoSanchez (changes requested):

splitio/storage/pluggable.py
chillaq: will fix

splitio/storage/pluggable.py
chillaq: will remove

splitio/storage/pluggable.py
chillaq: will fix

splitio/storage/pluggable.py
chillaq: will fix

splitio/storage/pluggable.py
chillaq: will fix

tests/storage/test_pluggable.py
chillaq: fixed

splitio/storage/pluggable.py
chillaq: fixed

EmilianoSanchez (approved):

PR #315, opened 2023-02-23
Added SegmentStorage class with tests

chillaq:
# Python SDK

## Extra Notes

### DIFF
--- a/splitio/storage/pluggable.py
+++ b/splitio/storage/pluggable.py
@@ -315,7 +315,7 @@ class PluggableSegmentStorage(SegmentStorage):
 #            _LOGGER.debug('Error: ', exc_info=True)
 #            return None
 
-    def segment_contains_key(self, segment_name, key):
+    def segment_contains(self, segment_name, key):
         """
         Check if segment contains a key
 
@@ -333,20 +333,6 @@ class PluggableSegmentStorage(SegmentStorage):
             _LOGGER.error('Error checking segment key')
             _LOGGER.debug('Error: ', exc_info=True)
             return None
-
-    def segment_contains(self, segment_name, key):
-        """
-        Check if segment contains a key, added for backward compatibility as its implemented in Redis
-
-        :param segment_name: segment name
-        :type segment_name: str
-        :param key: key
-        :type key: str
-
-        :return: True if found, otherwise False
-        :rtype: bool
-        """
-        self.segment_contains_key(segment_name, key)
 
     def get_segment_keys_count(self):
         """
--- a/splitio/storage/pluggable.py
+++ b/splitio/storage/pluggable.py
@@ -240,7 +240,7 @@ class PluggableSegmentStorage(SegmentStorage):
             if to_remove is not None:
                 self._pluggable_adapter.remove_items(self._prefix.format(segment_name=segment_name), to_remove)
             if change_number is not None:
-                self.set_change_number(segment_name, change_number)
+                self._pluggable_adapter.set(self._segment_till_prefix.format(segment_name=segment_name), change_number)
         except Exception:
             _LOGGER.error('Error updating segment storage')
             _LOGGER.debug('Error: ', exc_info=True)
@@ -254,11 +254,13 @@ class PluggableSegmentStorage(SegmentStorage):
         :param change_number: change number
         :type segment_name: int
         """
-        try:
-            self._pluggable_adapter.set(self._segment_till_prefix.format(segment_name=segment_name), change_number)
-        except Exception:
-            _LOGGER.error('Error updating segment change number')
-            _LOGGER.debug('Error: ', exc_info=True)
+        pass
+        # TODO: To be added when producer mode is aupported
+#        try:
+#            self._pluggable_adapter.set(self._segment_till_prefix.format(segment_name=segment_name), change_number)
+#        except Exception:
+#            _LOGGER.error('Error updating segment change number')
+#            _LOGGER.debug('Error: ', exc_info=True)
 
     def get_change_number(self, segment_name):
         """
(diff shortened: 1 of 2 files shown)


splitio/storage/pluggable.py
```diff
@@ -207,3 +207,170 @@ def put(self, split):
         if existing_split is not None and existing_split.traffic_type_name != split.traffic_type_name:
             self._increase_traffic_type_count(split.traffic_type_name)
             self._decrease_traffic_type_count(existing_split.traffic_type_name)
+
+
+class PluggableSegmentStorage(SegmentStorage):
+    """Pluggable implementation of segment storage."""
+    _SEGMENT_NAME_LEMNGTH = 14
```
EmilianoSanchez, line 214: typo: `_SEGMENT_NAME_LENGTH`

splitio/storage/pluggable.py
```diff
@@ -207,3 +207,170 @@ def put(self, split):
...
+
+        :param segment_name: Name of the segment to update.
+        :type segment_name: str
+        :param to_add: Set of members to add to the segment.
+        :type to_add: set
+        :param to_remove: List of members to remove from the segment.
+        :type to_remove: Set
+        """
+        try:
+            self._pluggable_adapter.add_items(self._prefix.format(segment_name=segment_name), to_add)
```
EmilianoSanchez, line 238: Should we validate that `to_add` is not None and empty before calling `self._pluggable_adapter.add_items`? or `_pluggable_adapter.add_items` handles that case ?

Same for `remove_items`.

splitio/storage/pluggable.py
```diff
@@ -207,3 +207,170 @@ def put(self, split):
...
+            self._pluggable_adapter.remove_items(self._prefix.format(segment_name=segment_name), to_remove)
+            if change_number is not None:
+                self._pluggable_adapter.set(self._segment_till_prefix.format(segment_name=segment_name), change_number)
+        except Exception:
+            _LOGGER.error('Error updating segment storage')
+            _LOGGER.debug('Error: ', exc_info=True)
+
+    def set_change_number(self, segment_name, change_number):
```
EmilianoSanchez, line 246: You can replace line 241 by calling `set_change_number`. Actually, I think you can also remove `set_change_number` method, if the change number is only update in line 241.

splitio/storage/pluggable.py
```diff
@@ -207,3 +207,170 @@ def put(self, split):
...
+            for key in self._pluggable_adapter.get_keys_by_prefix(self._prefix[:-self._SEGMENT_NAME_LEMNGTH]):
+                if key[-self._TILL_LENGTH:] != 'till':
+                    keys.append(key[len(self._prefix[:-self._SEGMENT_NAME_LEMNGTH]):])
+            return keys
+        except Exception:
+            _LOGGER.error('Error getting segments')
+            _LOGGER.debug('Error: ', exc_info=True)
+            return None
+
+    def get_keys(self, segment_name):
```
EmilianoSanchez, line 296: I think we don't need `get_keys`, `get_segments_count` and `get_segments_keys_count` for this storage (idem for `RedisSegmentStorage`), because this data is not being sent by telemetry in consumer mode. I must check and update the spec.

Maybe, we can add a `TODO` comment on these methods to consider removing them before releasing the features, if they are actually not used anywhere in the code.

splitio/storage/pluggable.py
```diff
@@ -207,3 +207,170 @@ def put(self, split):
...
+        :rtype: int
+        """
+        try:
+            return sum([self._pluggable_adapter.get_items_count(key) for key in self._pluggable_adapter.get_keys_by_prefix(self._prefix)])
+        except Exception:
+            _LOGGER.error('Error getting segment keys')
+            _LOGGER.debug('Error: ', exc_info=True)
+            return None
+
+    def get(self, segment_name):
```
EmilianoSanchez, line 346: Do we need a `get` and `put` method? 
Up to you, consider keeping or removing them eventually if they are not used.

EmilianoSanchez (changes requested):

splitio/storage/pluggable.py
chillaq: will fix

splitio/storage/pluggable.py
chillaq: I will reuse set_change_number, it is in the spec, so we need to keep it.

splitio/storage/pluggable.py
chillaq: will do, although this method will only be called from SDK engine so there is always a valid array variable passed, but it does not hurt to add validation.

splitio/storage/pluggable.py
chillaq: The InMemory implementation has get and put, that is why I added them.

splitio/storage/pluggable.py
chillaq: InMemory has get_segments_count and get_segments_keys_count, so I think we need to keep them, I don't see get_keys though, so will commented it out.

splitio/storage/pluggable.py
```diff
@@ -284,35 +288,55 @@ def get_segment_names(self):
...
+#        :rtype: str[]
+#        """
+#        try:
+#            return list(self._pluggable_adapter.get(self._prefix.format(segment_name=segment_name)))
+#        except Exception:
+#            _LOGGER.error('Error getting segments keys')
+#            _LOGGER.debug('Error: ', exc_info=True)
+#            return None
+
+    def segment_contains_key(self, segment_name, key):
```
EmilianoSanchez, line 318: Lets keep only `segment_contains`, and remove `segment_contains_key` for compatibility with redis

EmilianoSanchez (approved):

splitio/storage/pluggable.py
chillaq: done

PR #317, opened 2023-02-28
added impressions pluggable storage and tests

chillaq:
# Python SDK

## Extra Notes

### DIFF
--- a/splitio/storage/pluggable.py
+++ b/splitio/storage/pluggable.py
@@ -486,7 +486,11 @@ class PluggableImpressionsStorage(ImpressionStorage):
         :type sdk_metadata: splitio.client.util.SdkMetadata
         """
         self._pluggable_adapter = pluggable_adapter
-        self._sdk_metadata = sdk_metadata
+        self._sdk_metadata = {
+                                's': sdk_metadata.sdk_version,
+                                'n': sdk_metadata.instance_name,
+                                'i': sdk_metadata.instance_ip,
+                            }
         self._impressions_queue_key = 'SPLITIO.impressions'
         if prefix is not None:
             self._impressions_queue_key = prefix + "." + self._impressions_queue_key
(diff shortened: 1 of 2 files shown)


splitio/storage/pluggable.py
```diff
@@ -470,3 +472,87 @@ def put(self, segment):
...
+                        'i': self._sdk_metadata.instance_ip,
+                    },
+                    'i': {  # IMPRESSION PORTION
+                        'k': impression.matching_key,
+                        'b': impression.bucketing_key,
+                        'f': impression.feature_name,
+                        't': impression.treatment,
+                        'r': impression.label,
+                        'c': impression.change_number,
+                        'm': impression.time,
```
EmilianoSanchez, line 520: missing field `pt: impression.pt` (a.k.a previousTime)

splitio/storage/pluggable.py
```diff
@@ -470,3 +472,87 @@ def put(self, segment):
...
+        :type impressions: splitio.models.impressions.Impression
+
+        :return: Processed impressions.
+        :rtype: list[splitio.models.impressions.Impression]
+        """
+        bulk_impressions = []
+        for impression in impressions:
+            if isinstance(impression, Impression):
+                to_store = {
+                    'm': {  # METADATA PORTION
```
EmilianoSanchez, line 508: Since `self._sdk_metadata` is an immutable object, we can create the `'m'` object only once in the `__init__` function.

For example, instead of doing `self._sdk_metadata = sdk_metadata` we can do something like 
```
self._sdk_metadata = {  # METADATA PORTION
                        's': sdk_metadata.sdk_version,
                        'n': sdk_metadata.instance_name,
                        'i': sdk_metadata.instance_ip,
}
```
and then assign to `m` the object: `'m': self._sdk_metadata`

splitio/storage/pluggable.py
chillaq: I don't see this field in the spec, also Redis implementation does not have it.

splitio/storage/pluggable.py
chillaq: will do

splitio/storage/pluggable.py
EmilianoSanchez: You are right. I will add it to the spec. 
We can add the field in another PR.

EmilianoSanchez (approved):

PR #321, opened 2023-03-02
Added telemetry pluggable storage with tests

chillaq:
# Python SDK

## Extra Notes

### DIFF
--- a/splitio/storage/pluggable.py
+++ b/splitio/storage/pluggable.py
@@ -2,10 +2,12 @@
 
 import logging
 import json
+import threading
 
 from splitio.models import splits, segments
 from splitio.models.impressions import Impression
-from splitio.storage import SplitStorage, SegmentStorage, ImpressionStorage, EventStorage
+from splitio.models.telemetry import MethodExceptions, MethodLatencies, TelemetryConfig, MAX_TAGS
+from splitio.storage import SplitStorage, SegmentStorage, ImpressionStorage, EventStorage, TelemetryStorage
 
 _LOGGER = logging.getLogger(__name__)
 
@@ -15,7 +17,14 @@ class PluggableSplitStorage(SplitStorage):
     _SPLIT_NAME_LENGTH = 12
 
     def __init__(self, pluggable_adapter, prefix=None):
-        """Constructor."""
+        """
+        Class constructor.
+
+        :param pluggable_adapter: Storage client or compliant interface.
+        :type pluggable_adapter: TBD
+        :param prefix: optional, prefix to storage keys
+        :type prefix: str
+        """
         self._pluggable_adapter = pluggable_adapter
         self._prefix = "SPLITIO.split.{split_name}"
         self._traffic_type_prefix = "SPLITIO.trafficType.{traffic_type_name}"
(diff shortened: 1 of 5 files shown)


splitio/storage/pluggable.py
```diff
@@ -657,3 +678,167 @@ def clear(self):
...
+        """
+        record latency data
+
+        :param method: method name
+        :type method: string
+        :param latency: latency
+        :type latency: int64
+        """
+        self._method_latencies.add_latency(method, latency)
+        latencies = self._method_latencies.pop_all()['methodLatencies']
```
mmelograno, line 784: why we need to pop things, you need to increment one bucket directly into the storage (same logic as you've added into exceptions)

mmelograno (changes requested):

splitio/storage/pluggable.py
chillaq: This is an intermediary step to get the correct formatted latency bucket, since the logic is already in Telemetry classes, you see that we add the method, then get the bucket immediately.

chillaq: The latencies have mappings of values, they are not direct values like exceptions

mmelograno: why not doing this:
```
bucket = get_latency_bucket_index(latency)
latency_key = self._telemetry_latencies_key + '::' + self._sdk_metadata + '/' + method.value + '/' + str(bucket)
result = self._pluggable_adapter.increment(latency_key, 1)
```

mmelograno: could you provide me an example of a `latency_key` created?

mmelograno (changes requested):

splitio/storage/pluggable.py
```diff
@@ -657,3 +678,167 @@ def clear(self):
...
+        values = latencies[method.value]
+        total_keys = 0
+        bucket_number = 0
+        for bucket in values:
+            if bucket > 0:
+                latency_key = self._telemetry_latencies_key + '::' + self._sdk_metadata + '/' + method.value + '/' + str(bucket_number)
+                result = self._pluggable_adapter.increment(latency_key, bucket)
+                self.expire_keys(latency_key, self._TELEMETRY_KEY_DEFAULT_TTL, 1, result)
+                total_keys += 1
+            bucket_number = bucket_number + 0
```
mmelograno, line 794: ???????

mmelograno (changes requested):

splitio/storage/pluggable.py
chillaq: Using inmemory dict for keys, example below:
{'SPLITIO.telemetry.latencies::python-9.4.0/1.1.1.1/ip1.1.1.1/0': 1}

your code looks good, will test it

splitio/storage/pluggable.py
chillaq: This looks like a bug in redis class, using your code it is not needed though, will apply it to redis class as well

splitio/storage/pluggable.py
```diff
@@ -657,3 +678,159 @@ def clear(self):
...
+
+    def record_bur_time_out(self):
+        """Not implemented"""
+        pass
+
+    def record_impression_stats(self, data_type, count):
+        """Not implemented"""
+        pass
+
+    def expire_latency_keys(self, total_keys, inserted):
```
EmilianoSanchez, line 811: is this method used? If not, we can remove it (same for `RedisTelemetryStorage`)

chillaq: Yes it is used in the recorder class

PR #322, opened 2023-03-06
added validation and test

chillaq:
# Python SDK

## Extra Notes

### DIFF
--- a/splitio/client/input_validator.py
+++ b/splitio/client/input_validator.py
@@ -3,6 +3,7 @@ from numbers import Number
 import logging
 import re
 import math
+import inspect
 
 from splitio.api import APIException
 from splitio.api.commons import FetchOptions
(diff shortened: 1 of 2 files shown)


splitio/client/input_validator.py
```diff
@@ -517,3 +518,46 @@ def valid_properties(properties):
...
+    for exp_method in expected_methods:
+        method_found = False
+        get_method_args = set()
+        for method in methods:
+            if exp_method == method[0]:
+                method_found = True
+                get_method_args = inspect.signature(method[1]).parameters
+                break
+        if not method_found:
+            _LOGGER.error("Pluggable adapter does not have required method: %s" % exp_method)
```
EmilianoSanchez, line 558: In this validation, we are throwing an error in JS, because it is like a kind of compile-time error of hard typed languages.

splitio/client/input_validator.py
```diff
@@ -517,3 +518,46 @@ def valid_properties(properties):
...
+    """
+    Check if pluggable adapter contains the expected method signature
+
+    :param config: config parameters
+    :type config: Dict
+
+    :return: True if no issue found otherwise False
+    :rtype: bool
+    """
+    if config.get('storageType') != 'PLUGGABLE':
```
EmilianoSanchez, line 532: besides `storageType` and `wrapper`, should we also validate the `prefix` config option here?

splitio/client/input_validator.py
chillaq: Yeah, was planning to do that in the factory class

splitio/client/input_validator.py
chillaq: Sure, will add validation that it is string type.

EmilianoSanchez (approved):

PR #324, opened 2023-03-09
Pluggable integration tests

chillaq:
# Python SDK

## Extra Notes

### DIFF
--- a/tests/integration/test_pluggable_integration.py
+++ b/tests/integration/test_pluggable_integration.py
@@ -1,0 +1,247 @@
+"""Redis storage end to end tests."""
+#pylint: disable=no-self-use,protected-access,line-too-long,too-few-public-methods
+
+import json
+import os
+
+from splitio.client.util import get_metadata
+from splitio.models import splits, impressions, events
+from splitio.storage.pluggable import PluggableEventsStorage, PluggableImpressionsStorage, PluggableSegmentStorage, \
+    PluggableSplitStorage, PluggableTelemetryStorage
+from splitio.client.config import DEFAULT_CONFIG
+from tests.storage.test_pluggable import StorageMockAdapter
+
+class PluggableSplitStorageIntegrationTests(object):
+    """Pluggable Split storage e2e tests."""
+
+    def test_put_fetch(self):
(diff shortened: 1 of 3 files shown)


tests/integration/test_pluggable_integration.py
```diff
@@ -0,0 +1,247 @@
...
+        ])
+
+
+    def test_put_fetch_contains(self):
+        """Test storing and retrieving splits in pluggable."""
+        adapter = StorageMockAdapter()
+        try:
+            self._put_impressions(adapter, get_metadata({}))
+
+            imps = adapter.get('SPLITIO.impressions')
```
EmilianoSanchez, line 172: it should be `imps = adapter.pop_items('SPLITIO.impressions')`

For completeness, I would add `pop_items` to your wrapper interface, although you will not use it at the moment, because it is used by the producer method (`pop_many`) of impressions and events pluggable storages.

`pop_items` retrieves the items added via `push_items` operation. Together with `get_items_count`, these 3 operations are "list" operation types, equivalent to  `rpop`, `rpush` and `llen` in Redis respectively.

tests/integration/test_pluggable_integration.py
```diff
@@ -0,0 +1,247 @@
...
+
+    def test_put_fetch_contains_ip_address_disabled(self):
+        """Test storing and retrieving splits in pluggable."""
+        adapter = StorageMockAdapter()
+        try:
+            cfg = DEFAULT_CONFIG.copy()
+            cfg.update({'IPAddressesEnabled': False})
+            self._put_impressions(adapter, get_metadata(cfg))
+
+            imps = adapter.get('SPLITIO.impressions')
```
EmilianoSanchez, line 189: idem: `imps = adapter.pop_items('SPLITIO.impressions')`

EmilianoSanchez (changes requested):

tests/integration/test_pluggable_integration.py
chillaq: sure, will do

EmilianoSanchez (approved):

PR #341, opened 2023-05-05
Httpclient async lib

chillaq:
# Python SDK


splitio/api/client.py
```diff
@@ -3,22 +3,12 @@
...
-        Class constructor.
+from splitio.api import build_basic_headers, HttpClientException
 
-        :param message: Information on why this exception happened.
-        :type message: str
-        """
-        Exception.__init__(self, message)
+_LOGGER = logging.getLogger(__name__)
 
+HttpResponse = namedtuple('HttpResponse', ['status_code', 'body'])
```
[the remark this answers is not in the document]
chillaq: will do

splitio/api/client_async.py
```diff
@@ -0,0 +1,137 @@
...
+        :type extra_headers: dict
+
+        :return: Tuple of status_code & response text
+        :rtype: HttpResponse
+        """
+        headers = build_basic_headers(apikey)
+        if extra_headers is not None:
+            headers.update(extra_headers)
+        try:
+            async with aiohttp.ClientSession() as session:
```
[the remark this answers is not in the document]
chillaq: Will do

splitio/api/client.py
```diff
@@ -1,33 +1,59 @@
...
+
+    :param server: Server for whith the request is being made.
+    :type server: str
+    :param path: URL path to be appended to base host.
+    :type path: str
+
+    :return: A fully qualified URL.
+    :rtype: str
+    """
+    return urls[server] + path
```
mredolatti, line 34: https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urljoin

splitio/api/client.py
```diff
@@ -1,33 +1,59 @@
...
+
+def _construct_urls(sdk_url=None, events_url=None, auth_url=None, telemetry_url=None):
+    return {
+        'sdk': sdk_url if sdk_url is not None else SDK_URL,
+        'events': events_url if events_url is not None else EVENTS_URL,
+        'auth': auth_url if auth_url is not None else AUTH_URL,
+        'telemetry': telemetry_url if telemetry_url is not None else TELEMETRY_URL,
+    }
+
+class HttpClientBase(object):
```
mredolatti, line 44: this should be abstract, right?

splitio/api/commons.py
```diff
@@ -113,4 +115,41 @@ def build_fetch(change_number, fetch_options, metadata):
...
+        """Constructor."""
+        Exception.__init__(self, custom_message)
+        self._status_code = status_code if status_code else -1
+
+    @property
+    def status_code(self):
+        """Return HTTP status code."""
+        return self._status_code
+
+class HttpClientException(Exception):
```
mredolatti, line 133: this exception is only thrown by the HTTP clients, i think it belongs to that module

splitio/api/commons.py
```diff
@@ -113,4 +115,41 @@ def build_fetch(change_number, fetch_options, metadata):
...
+    def __init__(self, message):
+        """
+        Class constructor.
+
+        :param message: Information on why this exception happened.
+        :type message: str
+        """
+        Exception.__init__(self, message)
+
+def build_basic_headers(apikey):
```
mredolatti, line 145: same goes for this function, it should only be used by the HTTP client, so it's better off scoped to that module

splitio/api/commons.py
```diff
@@ -1,9 +1,11 @@
 """Commons module."""
+from collections import namedtuple
 from splitio.util.time import get_current_epoch_time_ms
 
 _CACHE_CONTROL = 'Cache-Control'
 _CACHE_CONTROL_NO_CACHE = 'no-cache'
 
+HttpResponse = namedtuple('HttpResponse', ['status_code', 'body'])
```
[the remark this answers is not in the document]
mredolatti: please add response headres and move this to the httpclient module

splitio/api/client.py
chillaq: will do

splitio/api/client.py
chillaq: Yep, will add abc lib

splitio/api/commons.py
chillaq: will move

splitio/api/commons.py
chillaq: will move

splitio/api/commons.py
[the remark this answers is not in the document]
chillaq: will move, but why do we need to add response headers? This is the original code

mredolatti (approved):

PR #358, opened 2023-05-09
update vars in engine.evaluator

chillaq:
# Python SDK

### DIFF
--- a/splitio/engine/evaluator.py
+++ b/splitio/engine/evaluator.py
@@ -13,25 +13,25 @@ _LOGGER = logging.getLogger(__name__)
 class Evaluator(object):  # pylint: disable=too-few-public-methods
     """Split Evaluator class."""
 
-    def __init__(self, split_storage, segment_storage, splitter):
+    def __init__(self, feature_flag_storage, segment_storage, splitter):
         """
         Construct a Evaluator instance.
 
-        :param split_storage: Split storage.
-        :type split_storage: splitio.storage.SplitStorage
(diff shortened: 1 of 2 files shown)


splitio/engine/evaluator.py
```diff
@@ -43,51 +43,51 @@ def _evaluate_treatment(self, feature, matching_key, bucketing_key, attributes,
         :param attributes: An optional dictionary of attributes
         :type attributes: dict
 
-        :param split: Split object
+        :param feature_flag: Split object
```
gthea, line 46: Double checking, is `Split` a type or should it also be changed?

splitio/engine/evaluator.py
```diff
@@ -128,24 +128,24 @@ def evaluate_features(self, features, matching_key, bucketing_key, attributes=No
...
         """
         Evaluate the feature considering the conditions.
 
         If there is a match, it will return the condition and the label.
         Otherwise, it will return (None, None)
 
-        :param split: The split for which to get the treatment
-        :type split: Split
+        :param feature_flag: The feature flag for which to get the treatment
+        :type feature_flag: Split
```
gthea, line 148: Double checking, is `Split` a type or should it be changed?

chillaq: It is a class, I put the task to change it in "Could" list

splitio/engine/evaluator.py
chillaq: It is a class, I put the task to change it in "Could" list

PR #364, opened 2023-05-10
updated apikey in splitio.api classes

chillaq:
# Python SDK


gthea (commented):
Q: I understand the previous name was `apikey`, but would it make sense to use underscores for the new one? As in `sdk_key`

chillaq:
> Q: I understand the previous name was `apikey`, but would it make sense to use underscores for the new one? As in `sdk_key`

Yeah, it can go either way, I wanted to keep the format as is.

gthea (approved):

PR #365, opened 2023-05-10
updated apikey in splitio.client classes

chillaq:
# Python SDK

### DIFF
--- a/splitio/client/client.py
+++ b/splitio/client/client.py
@@ -377,7 +377,7 @@ class Client(object):  # pylint: disable=too-many-instance-attributes
         start = get_current_epoch_time_ms()
         key = input_validator.validate_track_key(key)
         event_type = input_validator.validate_event_type(event_type)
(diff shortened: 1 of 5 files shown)


splitio/client/client.py
```diff
@@ -377,7 +377,7 @@ def track(self, key, traffic_type, event_type, value=None, properties=None):
         start = get_current_epoch_time_ms()
         key = input_validator.validate_track_key(key)
         event_type = input_validator.validate_event_type(event_type)
-        should_validate_existance = self.ready and self._factory._apikey != 'localhost'  # pylint: disable=protected-access
+        should_validate_existance = self.ready and self._factory._sdkkey != 'localhost'  # pylint: disable=protected-access
```
gthea, line 380: Since #364 uses `sdk_key` maybe we can update this one for consistency.

gthea (approved):

splitio/client/client.py
chillaq: Yep, updating it now.

PR #374, opened 2023-05-16
Added support for FF to push.parser class

chillaq:
# Python SDK


splitio/push/parser.py
```diff
@@ -326,9 +325,12 @@ def change_number(self):
 class SplitChangeUpdate(BaseUpdate):
     """Split Change notification."""
 
-    def __init__(self, channel, timestamp, change_number):
+    def __init__(self, channel, timestamp, change_number, previous_change_number, split_definition, compression):
```
mmelograno, line 328: let's start not using the word split at all

mmelograno (changes requested):

PR #378, opened 2023-05-19
Added e2e tests

chillaq:
# Python SDK

### DIFF
--- a/splitio/push/splitworker.py
+++ b/splitio/push/splitworker.py
@@ -8,6 +8,7 @@ import json
 from enum import Enum
 
 from splitio.models.splits import from_raw
+from splitio.push.parser import UpdateType
 
 _LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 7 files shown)


splitio/push/splitworker.py
```diff
@@ -63,16 +64,17 @@ def _run(self):
...
-                    try:
-                        self._feature_flag_storage.put(from_raw(json.loads(self._get_feature_flag_definition(event))))
-                        self._feature_flag_storage.set_change_number(event.change_number)
-                        continue
-                    except Exception as e:
-                        _LOGGER.error('Exception raised in updating feature flag')
-                        _LOGGER.debug(str(e))
-                        _LOGGER.debug('Exception information: ', exc_info=True)
-                        pass
+                if event.update_type == UpdateType.SPLIT_UPDATE:
```
mmelograno, line 67: why we need to check this type of event? I believe that those are the only events that are queued by the processor. If we need this check, I would suggest having the opposite and `pass` to avoid multiple if conditions

```
if event.update_type != UpdateType.SPLIT_UPDATE:
   do nothing
// rest of the code

File: tests/integration/test_streaming_e2e.py [hunk 2]
```diff
@@ -106,6 +109,21 @@ def test_happiness(self):
         assert factory.client().get_treatment('pindon', 'split2') == 'off'
         assert factory.client().get_treatment('maldo', 'split2') == 'on'
 
+        # test if changeNumber is missing
+#        split_changes = make_split_fast_change_event(4)
```
mmelograno, line 113: why commented code?

mmelograno (changes requested):

tests/integration/test_streaming_e2e.py
chillaq: The code does not check if the changeNumber is None, so this test will fail.
I wanted to check if we need to add checks to ignore the SPLIT_UPDATE and SPLIT_KILL events if the changeNumber value is missing.

mmelograno (approved):

PR #392, opened 2023-06-23
Added telemetry to iff update

chillaq:
# Python SDK

### DIFF
--- a/splitio/engine/telemetry.py
+++ b/splitio/engine/telemetry.py
@@ -139,9 +139,9 @@ class TelemetryRuntimeProducer(object):
         """Record session length."""
         self._telemetry_storage.record_session_length(session)
 
-    def record_update_from_sse(self):
-        """Record session length."""
-        self._telemetry_storage.record_update_from_sse()
+    def record_update_from_sse(self, event):
+        """Record update from sse."""
+        self._telemetry_storage.record_update_from_sse(event)
 
 class TelemetryStorageConsumer(object):
     """Telemetry storage consumer class."""
@@ -275,9 +275,9 @@ class TelemetryRuntimeConsumer(object):
         """Get and reset streaming events."""
         return self._telemetry_storage.pop_streaming_events()
 
-    def pop_update_from_sse(self):
+    def pop_update_from_sse(self, event):
         """Get and reset update from sse."""
-        return self._telemetry_storage.pop_update_from_sse()
+        return self._telemetry_storage.pop_update_from_sse(event)
 
     def get_session_length(self):
         """Get session length"""
--- a/splitio/engine/telemetry.py
+++ b/splitio/engine/telemetry.py
@@ -6,7 +6,7 @@ import logging
 _LOGGER = logging.getLogger(__name__)
 
 from  splitio.storage.inmemmory import InMemoryTelemetryStorage
-from splitio.models.telemetry import CounterConstants
+from splitio.models.telemetry import CounterConstants, UpdateFromSSE
 
 class TelemetryStorageProducer(object):
     """Telemetry storage producer class."""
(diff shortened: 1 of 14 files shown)


splitio/engine/telemetry.py
```diff
@@ -139,6 +139,10 @@ def record_session_length(self, session):
         """Record session length."""
         self._telemetry_storage.record_session_length(session)
 
+    def record_update_from_sse(self):
+        """Record session length."""
```
mmelograno, line 143: error doc

splitio/models/telemetry.py
```diff
@@ -519,6 +520,14 @@ def record_events_value(self, resource, value):
             else:
                 return
 
+    def record_update_from_sse(self):
+        """
+        Increament the update from sse resource by one.
```
mmelograno, line 525: typo

splitio/models/telemetry.py
```diff
@@ -604,6 +613,18 @@ def pop_token_refreshes(self):
             self._token_refreshes = 0
             return token_refreshes
 
+    def pop_update_from_sse(self):
+        """
+        Pop update from sse
+
+        :return: token refreshes value
```
mmelograno, line 620: error doc

mmelograno (changes requested):
You are not supporting multiple increments if that is the case in the future

splitio/engine/telemetry.py
chillaq: will fix

splitio/push/splitworker.py
```diff
@@ -75,6 +76,8 @@ def _run(self):
                         new_split = from_raw(json.loads(self._get_feature_flag_definition(event)))
                         if new_split.status == Status.ACTIVE:
                             self._feature_flag_storage.put(new_split)
+                            self._telemetry_runtime_producer.record_update_from_sse()
```
sanzmauro, line 79: you must to record the telemetry outside of the `if else`, because you are updating the storage in both scenarios.

splitio/push/segmentworker.py
```diff
@@ -11,7 +11,7 @@ class SegmentWorker(object):
 
     _centinel = object()
 
-    def __init__(self, synchronize_segment, segment_queue):
+    def __init__(self, synchronize_segment, segment_queue, telemetry_runtime_producer):
```
sanzmauro, line 14: why is this change? you are not using telemetry here.

splitio/models/telemetry.py
chillaq: will fix

splitio/models/telemetry.py
chillaq: will fix

splitio/models/telemetry.py
```diff
@@ -519,6 +520,14 @@ def record_events_value(self, resource, value):
             else:
                 return
 
+    def record_update_from_sse(self):
```
sanzmauro, line 523: you must to allow expansion for this telemetry because if in a future we starting to get `SPLIT_KILL` OR `SEGMENT_UPDATES` notifications with the definition inside we will want to track this telemetry too with the same method/storage.

sanzmauro (changes requested):

splitio/push/splitworker.py
chillaq: will fix

splitio/push/segmentworker.py
chillaq: ah yes, will remove

splitio/models/telemetry.py
chillaq: yes, will do

sanzmauro (changes requested):
Hey Bilal, what about this method? 
[pop_formatted_stats(self)](https://github.com/splitio/python-client/blob/4194be8f9a107aa3827d9e0865dd625d7dd8db76/splitio/engine/telemetry.py#L286C9-L286C35)

We should add this new property here. What do you think ?

chillaq:
> Hey Bilal, what about this method? [pop_formatted_stats(self)](https://github.com/splitio/python-client/blob/4194be8f9a107aa3827d9e0865dd625d7dd8db76/splitio/engine/telemetry.py#L286C9-L286C35)
> 
> We should add this new property here. What do you think ?

You are right, its in the sync.telemetry, I will move it to engine.telemetry

sanzmauro (approved):

PR #396, opened 2023-06-30
Async push manager

chillaq:
# Python SDK


splitio/push/sse.py
```diff
@@ -219,26 +219,11 @@ async def start(self, url, extra_headers=None):  # pylint:disable=protected-acce
             await self._conn.close()
             self._conn = None  # clear so it can be started again
             _LOGGER.debug("Existing SSEClient")
-            return
+        return
 
-    async def shutdown(self):
+    def shutdown(self):
```
mredolatti, line 224: what's the point of this? if it's just for the sake of interface compliance, i don't think it's worth it. either remove `shutdown()` from the interface spec or drop the interface altogether. sync & async implementations will never be interchangeable, so there's not much point in keeping them

splitio/push/manager.py
```diff
@@ -247,3 +289,248 @@ def _handle_connection_end(self):
...
+            if first_event.event == SSE_EVENT_ERROR:
+                raise(Exception("could not start SSE session"))
+
+            _LOGGER.debug("connected to streaming, scheduling next refresh")
+            self._token_task = asyncio.get_running_loop().create_task(self._token_refresh(token))
+            await self._handle_connection_ready()
+            await self._telemetry_runtime_producer.record_streaming_event((StreamingEventTypes.CONNECTION_ESTABLISHED, 0,  get_current_epoch_time_ms()))
+            await self._consume_events(events_source)
+        finally:
+            self._running = False
```
mredolatti, line 443: - how does all this code deal with exceptions? what happens if fetching the token throws?
- looking at splitsse.py it doesn't look like SSE_EVENT_ERROR is ever forwarded (https://github.com/splitio/python-client/pull/396/files#diff-acd0cce8988a1ea172a3377fb16fca0aa5c1c19dab842fa1eaeabc07a1e5ab00R205-R207). do we need to check for that here? In case the first event is an error, shouldn't we handle it?
- we should cover that any line in the function throwing an exception is gracefully handled (mapped to a retryable OR a non-retryable error)
- we should not let exceptions escape tasks that are directly scheduled in the event loop, since we lose control over them, and all we'll get is a log on the client app, without being able to react or handle the error properly

splitio/push/splitsse.py
```diff
@@ -200,16 +200,16 @@ async def start(self, token):
         self.status = SplitSSEClient._Status.CONNECTING
         url = self._build_url(token)
         try:
-            sse_events_task = self._client.start(url, extra_headers=self._metadata)
-            first_event = await anext(sse_events_task)
+            self.sse_events_task = self._client.start(url, extra_headers=self._metadata)
```
mredolatti, line 203: if the variable needs to be scoped to the instance and not the method, lets initialize to `None`, on the `__init__` method (and check it to have a value before using it on `shutdown()`

splitio/push/splitsse.py
```diff
@@ -225,5 +225,8 @@ async def stop(self, blocking=False, timeout=None):
         if self.status == SplitSSEClient._Status.IDLE:
             _LOGGER.warning('sse already closed. ignoring')
             return
-        await self._client.shutdown()
+        temp_task = asyncio.get_running_loop().create_task(anext(self.sse_events_task))
+        temp_task.cancel()
+        with suppress(asyncio.CancelledError):
+            await temp_task
```
mredolatti, line 231: what's the purpose of this?

mredolatti (changes requested):

splitio/push/manager.py
```diff
@@ -2,22 +2,62 @@
...
+    @abc.abstractmethod
+    def start(self):
+        """Start a new connection if not already running."""
 
-class PushManager(object):  # pylint:disable=too-many-instance-attributes
+    @abc.abstractmethod
+    def stop(self, blocking=False):
+        """Stop the current ongoing connection."""
+
+    def _get_parsed_event(self, event):
```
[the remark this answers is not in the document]
chillaq: will do

splitio/push/manager.py
```diff
@@ -247,3 +289,248 @@ def _handle_connection_end(self):
...
+
+    async def start(self):
+        """Start a new connection if not already running."""
+        if self._running:
+            _LOGGER.warning('Push manager already has a connection running. Ignoring')
+            return
+
+        try:
+            self._running_task = asyncio.get_running_loop().create_task(self._trigger_connection_flow())
+        except Exception as e:
```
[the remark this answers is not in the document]
chillaq: will remove the try block

splitio/push/sse.py
chillaq: I do agree its not worth it, but all SDKs are following the same spec, if we want to keep consistency across all, we should follow the same interface just for that.

splitio/push/splitsse.py
chillaq: That is the workaround to fix the task still pending error when script is done, we basically wait for task cancellation to finish

splitio/push/manager.py
chillaq: * The original code only capture exception for self._get_auth_token(), the logic is in that call, if it fails it will set the status to Status.PUSH_RETRYABLE_ERROR
* The threading code us setting splitsse status to SplitSSEClient._Status.ERRORED if first event is error, this is missing from async, I will add it. I will rework that logic between splitsse and manager

chillaq: in the threading push manager classes, if the first event is error, I only see the status is set to SplitSSEClient._Status.ERRORED, but nothing is checking that status, is it a retryable or non-retryable error?

mredolatti (approved):

PR #405, opened 2023-07-12
added telemetr async model

chillaq:
# Python SDK

### DIFF
--- a/splitio/models/telemetry.py
+++ b/splitio/models/telemetry.py
@@ -180,7 +180,8 @@ class MethodLatencies(MethodLatenciesBase):
     def __init__(self):
         """Constructor"""
         self._lock = threading.RLock()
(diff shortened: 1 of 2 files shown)


splitio/models/telemetry.py
```diff
@@ -650,12 +1225,51 @@ def time(self):
         """
         return self._time
 
+class StreamingEventsAsync(object):
```
EmilianoSanchez, line 1228: No need for a `StreamingEventsBase interface here`, right?

EmilianoSanchez (approved):

splitio/models/telemetry.py
chillaq: Nope, there is no shared code or function between threading and asyncio classes

PR #408, opened 2023-07-14
moved telemetry call to api.client for async

chillaq:
# Python SDK

### DIFF
--- a/splitio/api/client.py
+++ b/splitio/api/client.py
@@ -5,6 +5,7 @@ import urllib
 import abc
 
 from splitio.optional.loaders import aiohttp
+from splitio.util.time import get_current_epoch_time_ms
 
 SDK_URL = 'https://sdk.split.io/api'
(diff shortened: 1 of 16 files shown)


splitio/api/client.py
```diff
@@ -212,6 +249,7 @@ async def get(self, server, path, apikey, query=None, extra_headers=None):  # py
                 timeout=self._timeout
             ) as response:
                 body = await response.text()
+                await self._record_telemetry(response.status, get_current_epoch_time_ms() - start)
```
EmilianoSanchez, line 252: Is it `response.status` or should be `response.status_code` ? In case of the latter, check other occurrences of `response.status`

EmilianoSanchez (approved):

splitio/api/client.py
chillaq: Yeah, this is a difference between aiohttp (for asyncio) and requests library for threading.

PR #412, opened 2023-07-17
added async recorder classes

chillaq:
# Python SDK

### DIFF
--- a/splitio/recorder/recorder.py
+++ b/splitio/recorder/recorder.py
@@ -87,6 +87,56 @@ class StandardRecorder(StatsRecorder):
         return self._event_sotrage.put(event)
 
 
+class StandardRecorderAsync(StatsRecorder):
+    """StandardRecorder async class."""
+
+    def __init__(self, impressions_manager, event_storage, impression_storage, telemetry_evaluation_producer):
+        """
+        Class constructor.
+
+        :param impressions_manager: impression manager instance
+        :type impressions_manager: splitio.engine.impressions.Manager
+        :param event_storage: event storage instance
+        :type event_storage: splitio.storage.EventStorage
+        :param impression_storage: impression storage instance
+        :type impression_storage: splitio.storage.ImpressionStorage
+        """
+        self._impressions_manager = impressions_manager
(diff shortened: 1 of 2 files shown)


splitio/recorder/recorder.py
```diff
@@ -87,6 +87,56 @@ def record_track_stats(self, event, latency):
...
+
+        :param impressions: impressions generated for each evaluation performed
+        :type impressions: array
+        :param latency: time took for doing evaluation
+        :type latency: int
+        :param operation: operation type
+        :type operation: str
+        """
+        try:
+            if method_name is not None:
```
EmilianoSanchez, line 121: do you need two params (method_name and operation), or can it be done with only one?
E.g.:
```
if operation is not None:
    await self._telemetry_evaluation_producer.record_latency(operation, latency)
```

splitio/recorder/recorder.py
```diff
@@ -87,6 +87,56 @@ def record_track_stats(self, event, latency):
...
+                await self._telemetry_evaluation_producer.record_latency(operation, latency)
+            impressions = self._impressions_manager.process_impressions(impressions)
+            await self._impression_storage.put(impressions)
+        except Exception:  # pylint: disable=broad-except
```
EmilianoSanchez, line 125: should you handle exception here or not, like in `record_treatment_stats`?

EmilianoSanchez (approved):
LGTM, but there are some minor comments to check

splitio/recorder/recorder.py
chillaq: method_name is optional from the original code, but operation is required. So the if has to move to the previous call.
This is the original code BTW.

splitio/recorder/recorder.py
chillaq: Did you mean record_track_stats call? this call returns True or False only, the called function does have try/catch block in case there is an exception.

EmilianoSanchez: Sorry, I meant `record_track_stats`

EmilianoSanchez (approved):

PR #419, opened 2023-07-19
Async engine unique keys tracker

chillaq:
# Python SDK

### DIFF
--- a/splitio/engine/impressions/unique_keys_tracker.py
+++ b/splitio/engine/impressions/unique_keys_tracker.py
@@ -7,8 +7,8 @@ from splitio.optional.loaders import asyncio
 
 _LOGGER = logging.getLogger(__name__)
 
-class BaseUniqueKeysTracker(object, metaclass=abc.ABCMeta):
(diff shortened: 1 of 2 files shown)


splitio/engine/impressions/unique_keys_tracker.py
```diff
@@ -1,22 +1,46 @@
...
+    def _add_or_update(self, feature_flag_name, key):
+        """
+        Add the feature_name+key to both bloom filter and dictionary.
+
+        :param feature_flag_name: feature flag name associated with the key
+        :type feature_flag_name: str
+        :param key: key to be added to MTK list
+        :type key: int
+        """
+        if feature_flag_name not in self._cache:
```
EmilianoSanchez, line 38: Do you need a `with self._lock:` here?

EmilianoSanchez (approved):

splitio/engine/impressions/unique_keys_tracker.py
chillaq: Not needed here since the calling line is within a self._lock block

PR #422, opened 2023-07-21
added impressions and impressions count sync classes

chillaq:
# Python SDK

### DIFF
--- a/splitio/sync/impression.py
+++ b/splitio/sync/impression.py
@@ -2,11 +2,13 @@ import logging
 import queue
 
 from splitio.api import APIException
+from splitio.optional.loaders import asyncio
 
 _LOGGER = logging.getLogger(__name__)
 
 
 class ImpressionSynchronizer(object):
+    """Impressions synchronizer class."""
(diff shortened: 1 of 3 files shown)


splitio/sync/impression.py
```diff
@@ -95,3 +97,95 @@ def synchronize_counters(self):
...
+        self._bulk_size = bulk_size
+        self._failed = asyncio.Queue()
+
+    async def _get_failed(self):
+        """Return up to <BULK_SIZE> impressions stored in the failed impressions queue."""
+        imps = []
+        count = 0
+        while count < self._bulk_size and self._failed.qsize() > 0:
+            try:
+                imps.append(await self._failed.get())
```
EmilianoSanchez, line 127: Check if it should be `imps.append(await self._failed.get(False))`

EmilianoSanchez (approved):

splitio/sync/impression.py
chillaq: The False is related to Threading Queue.Queue class for not blocking
asyncio.Queue does not have it

PR #424, opened 2023-07-24
Added workerpool and sync.segment async classes

chillaq:
# Python SDK

### DIFF
--- a/splitio/tasks/util/workerpool.py
+++ b/splitio/tasks/util/workerpool.py
@@ -4,8 +4,10 @@ import logging
 from threading import Thread, Event
 import queue
 
+from splitio.optional.loaders import asyncio
 
 _LOGGER = logging.getLogger(__name__)
+_ASYNC_SLEEP_SECONDS = 0.3
 
 
 class WorkerPool(object):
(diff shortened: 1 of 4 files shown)


splitio/tasks/util/workerpool.py
```diff
@@ -134,3 +134,99 @@ def _wait_workers_shutdown(self, event):
...
+        """process a single message."""
+        try:
+            await self._semaphore.acquire() # wait until "there's a free worker"
+            if self._aborted: # check in case the pool was shutdown while we were waiting for a worker
+                return
+            await self._handler(message._message)
+        except Exception:
+            _LOGGER.error("Something went wrong when processing message %s", message)
+            _LOGGER.debug('Original traceback: ', exc_info=True)
+            self._failed = True
```
mredolatti, line 176: i don't think `self._failed` will work, since it's signaling that something failed when processing one message, which may or may not be from the same batch, and since this is async, it can become somewhat unpredictable.

how do you feel about adding a `failed` property to the `TaskCompletionWrapper` instead (and maybe a method to `BatchCompletionHandler` as well, that checks the status of nested items to look for possible errors? that way we can track what failed (it can be a bool, the expcetion's message or both, follow your heart).

ideally we can consume the WP's interface like this:
```
       jobs = await self._worker_pool.submit_work(segment_names)
        if dont_wait:
            return True
        await jobs.await_completion()
        return jobs.all_successful()
```

mredolatti (approved):

splitio/tasks/util/workerpool.py
```diff
@@ -134,3 +134,98 @@ def _wait_workers_shutdown(self, event):
...
+
+
+class BatchCompletionWrapper:
+    """Batch completion class"""
+    def __init__(self, tasks):
+        self._tasks = tasks
+
+    async def await_completion(self):
+        await asyncio.gather(*[task.await_completion() for task in self._tasks])
+        for task in self._tasks:
```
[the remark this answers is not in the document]
chillaq: will update

splitio/tasks/util/workerpool.py
chillaq: fixed

PR #438, opened 2023-08-16
added redis syncrhonizer async class

chillaq:
# Python SDK

### DIFF
--- a/splitio/sync/synchronizer.py
+++ b/splitio/sync/synchronizer.py
@@ -550,9 +550,6 @@ class RedisSynchronizerAsync(RedisSynchronizerBase):
     async def _stop_periodic_data_recording(self):
         """
         Stop recorders.
-
-        :param blocking: flag to wait until tasks are stopped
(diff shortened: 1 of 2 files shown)


splitio/sync/synchronizer.py
```diff
@@ -488,6 +476,102 @@ def stop_periodic_fetching(self):
...
+        :type blocking: bool
+        """
+        _LOGGER.debug('Shutting down tasks.')
+        await self.stop_periodic_data_recording(blocking)
+
+    async def _stop_periodic_data_recording(self):
+        """
+        Stop recorders.
+
+        :param blocking: flag to wait until tasks are stopped
```
EmilianoSanchez, line 554: Remove `:param blocking`? Since it is not part of the method signature.

EmilianoSanchez (approved):

splitio/sync/synchronizer.py
chillaq: fixed

PR #445, opened 2023-08-22
Update split model

chillaq:
# Python SDK

### DIFF
--- a/splitio/models/splits.py
+++ b/splitio/models/splits.py
@@ -250,5 +250,5 @@ def from_raw(raw_split):
         traffic_allocation=raw_split.get('trafficAllocation'),
         traffic_allocation_seed=raw_split.get('trafficAllocationSeed'),
         configurations=raw_split.get('configurations'),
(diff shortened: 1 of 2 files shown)


splitio/models/splits.py
```diff
@@ -238,5 +249,6 @@ def from_raw(raw_split):
         raw_split.get('algo'),
         traffic_allocation=raw_split.get('trafficAllocation'),
         traffic_allocation_seed=raw_split.get('trafficAllocationSeed'),
-        configurations=raw_split.get('configurations')
+        configurations=raw_split.get('configurations'),
+        sets=raw_split['sets']
```
mmelograno, line 253: why `raw_split['sets'] and not check `raw_split.get('sets')`

mmelograno (changes requested):

splitio/models/splits.py
chillaq: will update it

mmelograno (approved):

PR #446, opened 2023-08-22
updated api.split and api.commons

chillaq:
# Python SDK

### DIFF
--- a/splitio/api/commons.py
+++ b/splitio/api/commons.py
@@ -125,5 +125,5 @@ def build_fetch(change_number, fetch_options, metadata):
     if fetch_options.change_number is not None:
         query['till'] = fetch_options.change_number
     if fetch_options.sets is not None:
-        query['sets'] = ','.join(fetch_options.sets)
+        query['sets'] = ','.join(sorted(fetch_options.sets))
     return query, extra_headers
--- a/splitio/api/commons.py
+++ b/splitio/api/commons.py
@@ -86,7 +86,7 @@ class FetchOptions(object):
 
     @property
     def sets(self):
-        """Return change number."""
+        """Return sets."""
         return self._sets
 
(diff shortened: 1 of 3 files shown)


mmelograno (changes requested):
this not supports ordering alphabetically which generates several fastly cached keys

chillaq:
Fixed

splitio/api/commons.py
```diff
@@ -125,5 +125,5 @@ def build_fetch(change_number, fetch_options, metadata):
     if fetch_options.change_number is not None:
         query['till'] = fetch_options.change_number
     if fetch_options.sets is not None:
-        query['sets'] = ','.join(fetch_options.sets)
+        query['sets'] = ','.join(sorted(fetch_options.sets))
```
mmelograno, line 128: correct me if I'm wrong but this seems to be executed every single time that the logic makes a fetch, if that so, this is not as performant as doing this when you set/create the FetchOptions

mmelograno (changes requested):

splitio/api/commons.py
chillaq: The FetchOptions object is created every time a sync is performed in sync.segment synchronize_segment() and sync.splits synchronize_splits(), so it would not make a difference to move the line there.
Unless we create the FetchOoptions object in the sync initializers, however, we still need to update the query sets for any changes, which also means looping through the sets.
Let me know how you want to proceed.

chillaq: Yeah, we still need to update the sets for any possible changes, in the sync class before calling the fetch api, which means looping through the sets, will update the PR

chillaq: PR is updated

mmelograno (approved):

PR #455, opened 2023-08-29
Added sync.split logic, updated storage.inmemory.split and sync.synch…

chillaq:
…ronizer classes

# Python SDK

### DIFF
--- a/splitio/storage/inmemmory.py
+++ b/splitio/storage/inmemmory.py
@@ -82,9 +82,9 @@ class InMemorySplitStorage(SplitStorage):
             self._increase_traffic_type_count(split.traffic_type_name)
             if split.sets is not None:
                 for flag_set in split.sets:
-                    if flag_set not in self._sets_feature_flag_map.keys() and self.config_flag_sets_used > 0:
-                        continue
                     if flag_set not in self._sets_feature_flag_map.keys():
+                        if self.config_flag_sets_used > 0:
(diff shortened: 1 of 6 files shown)


splitio/storage/inmemmory.py
```diff
@@ -114,7 +119,7 @@ def _remove_flag_sets(self, feature_flag):
         if feature_flag.sets is not None:
```
EmilianoSanchez, line 119: Should this be like:
```
        if self.config_flag_sets_used != 0 and feature_flag.sets is not None:
            for flag_set in feature_flag.sets:
                self._sets_feature_flag_map[flag_set].remove(feature_flag.name)
                if len(self._sets_feature_flag_map[flag_set]) == 0:
                    del self._sets_feature_flag_map[flag_set]
```
?
I.e., only remove feature flag names from the flag set map if `config_flag_sets_used != 0`. Otherwise, the user hasn't passed flag sets in the config, so there is no need to update the flag set map.

EmilianoSanchez (approved):

splitio/storage/inmemmory.py
EmilianoSanchez: Ignore my comment @chillaq . I now understand that we need to keep the flag set map in order to not limit the usage of `getTreamentByFlagset` if flag set filter is not provided.

PR #457, opened 2023-08-31
added client.manager test, minor fix in models.splits and updated con…

chillaq:
…fig param name

# Python SDK

### DIFF
--- a/splitio/models/splits.py
+++ b/splitio/models/splits.py
@@ -200,7 +200,7 @@ class Split(object):  # pylint: disable=too-many-instance-attributes
             list(set(part.treatment for cond in self.conditions for part in cond.partitions)),
             self.change_number,
             self._configurations if self._configurations is not None else {},
-            self._sets
+            self._sets if self._sets is not None else []
         )
 
     def local_kill(self, default_treatment, change_number):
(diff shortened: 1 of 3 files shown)


splitio/models/splits.py
```diff
@@ -200,7 +200,7 @@ def to_split_view(self):
             list(set(part.treatment for cond in self.conditions for part in cond.partitions)),
             self.change_number,
             self._configurations if self._configurations is not None else {},
-            self._sets
+            self._sets if self._sets is not None else []
```
gthea, line 203: 👍

tests/client/test_manager.py
```diff
@@ -55,3 +56,37 @@ def test_evaluations_before_running_post_fork(self, mocker):
...
+        self._verify_split(split_view)
+
+    def _verify_split(self, split):
+        assert split.name == 'some_name'
+        assert split.traffic_type == 'user'
+        assert split.killed == False
+        assert split.treatments == ['on', 'off']
+        assert split.change_number == 123
+        assert split.configs == {'on': '{"color": "blue", "size": 13}'}
+        assert split.sets == ['set1', 'set2']
```
gthea, line 92: Maybe a test to check that sets equals `[]` when the split has no sets could be added

gthea (approved):

tests/client/test_manager.py
chillaq: will do

PR #458, opened 2023-08-31
Added flagset support to factory

chillaq:
# Python SDK

### DIFF
--- a/splitio/client/factory.py
+++ b/splitio/client/factory.py
@@ -350,7 +350,7 @@ def _build_in_memory_factory(api_key, cfg, sdk_url=None, events_url=None,  # pyl
     }
 
     storages = {
-        'splits': InMemorySplitStorage(cfg['FlagSetsFilter'] if cfg['FlagSetsFilter'] is not None else []),
+        'splits': InMemorySplitStorage(cfg['flagSetsFilter'] if cfg['flagSetsFilter'] is not None else []),
         'segments': InMemorySegmentStorage(),
         'impressions': InMemoryImpressionStorage(cfg['impressionsQueueSize'], telemetry_runtime_producer),
         'events': InMemoryEventStorage(cfg['eventsQueueSize'], telemetry_runtime_producer),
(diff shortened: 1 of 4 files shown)


splitio/client/factory.py
```diff
@@ -350,7 +350,7 @@ def _build_in_memory_factory(api_key, cfg, sdk_url=None, events_url=None,  # pyl
     }
 
     storages = {
-        'splits': InMemorySplitStorage(),
+        'splits': InMemorySplitStorage(cfg['FlagSetsFilter'] if cfg['FlagSetsFilter'] is not None else []),
```
mmelograno, line 353: we should be consistent with cfg and use first letter lowercase

mmelograno (changes requested):

splitio/client/factory.py
chillaq: will fix

splitio/models/telemetry.py
```diff
@@ -41,6 +41,7 @@ class ConfigParams(Enum):
     EVENTS_QUEUE_SIZE = 'eventsQueueSize'
     IMPRESSIONS_MODE = 'impressionsMode'
     IMPRESSIONS_LISTENER = 'impressionListener'
+    FLAG_SETS = 'FlagSetsFilter'
```
mmelograno, line 44: should start with lowecase?

mmelograno (changes requested):

splitio/models/telemetry.py
chillaq: Correct,  the class (ConfigParams) should be lower case, will check and correct all the constant classes in Telemetry model.

splitio/models/telemetry.py
```diff
@@ -27,7 +27,7 @@ class CounterConstants(Enum):
     EVENTS_QUEUED = 'eventsQueued'
     EVENTS_DROPPED = 'eventsDropped'
 
-class ConfigParams(Enum):
+class _ConfigParams(Enum):
```
mmelograno, line 30: why classes with `_` at the begining?

mmelograno (changes requested):

splitio/models/telemetry.py
chillaq: The constants class is used only internal in telemetry model.

PR #464, opened 2023-09-06
updted storage.pluggable class

chillaq:
# Python SDK


splitio/storage/pluggable.py
```diff
@@ -26,51 +27,78 @@ def __init__(self, pluggable_adapter, prefix=None):
...
+            if sets_to_fetch == []:
+                return []
+
+            keys = [self._prefix(feature_flag_name) for feature_flag_name in sets_to_fetch]
+            result_sets = self._pluggable_adapter.get_many(keys)
+            return list(combine_valid_flag_sets(result_sets))
+        except Exception:
+            _LOGGER.error('Error fetching feature flag from storage')
+            _LOGGER.debug('Error: ', exc_info=True)
+            return None
```
EmilianoSanchez, line 99: Should it return `[]` for consistency with line 91? Or is `None` used to represent that the function had an error?

EmilianoSanchez (approved):

splitio/storage/pluggable.py
chillaq: Yep, I added a None check from client class, thanks.

PR #466, opened 2023-09-11
Fixed all tests

chillaq:
# Python SDK


mmelograno (approved):

splitio/models/flag_sets.py
```diff
@@ -0,0 +1,124 @@
...
+        """
+        if not self.should_filter:
+            return True
+        if not isinstance(flag_sets, set) or len(flag_sets) == 0:
+            return False
+        return any(self.flag_sets.intersection(flag_sets))
+
+
+class FlagSets(object):
+    """InMemory Flagsets storage."""
```
mmelograno, line 45: this is storage, not a model

mmelograno (changes requested):

splitio/models/flag_sets.py
chillaq: Will keep the model and move the storage.

PR #468, opened 2023-09-12
Push manager pr

chillaq:
# Python SDK


splitio/push/manager.py
```diff
@@ -143,7 +143,7 @@ def _trigger_connection_flow(self):
             self._feedback_loop.put(Status.PUSH_RETRYABLE_ERROR)
             return
 
-        if not token.push_enabled:
+        if token is None or not token.push_enabled:
```
mmelograno, line 146: I think we should change the logic of Token dto and instead of returning this
`    return None if push_enabled is None else Token(push_enabled, token, json.loads(decoded_token['x-ably-capability']), decoded_token['exp'], decoded_token['iat'])
`
return Token(False, whatever, whatever, whatever, whatever) is still a valid Token dto without Token payload. Then, check push_enabled instead of None

tests/push/test_manager.py
```diff
@@ -90,6 +90,28 @@ def new_start(*args, **kwargs):  # pylint: disable=unused-argument
         assert feedback_loop.get() == Status.PUSH_RETRYABLE_ERROR
         assert timer_mock.mock_calls == [mocker.call(0, Any())]
 
+    def test_empty_auth_respnse(self, mocker):
```
mmelograno, line 93: correct me if i'm wrong but this test seems to be equal to the one below, moreover you have typos

mmelograno (changes requested):

splitio/push/manager.py
chillaq: Will do

tests/push/test_manager.py
mredolatti: how about we update that function that fetches the token to either return a token (regardless of whether push is enabled or not) or throw an exception otherwise? That will help keep the API cleaner

mredolatti (changes requested):

mmelograno (approved):

mredolatti (approved):

PR #469, opened 2023-09-19
Added flagsets total and invalid count for telemetry config

chillaq:
# Python SDK

### DIFF
--- a/splitio/client/factory.py
+++ b/splitio/client/factory.py
@@ -69,6 +69,9 @@ _INSTANTIATED_FACTORIES = Counter()
 _INSTANTIATED_FACTORIES_LOCK = threading.RLock()
 _MIN_DEFAULT_DATA_SAMPLING_ALLOWED = 0.1  # 10%
 _MAX_RETRY_SYNC_ALL = 3
+_FLAG_SETS_LOCK = threading.RLock()
+_TOTAL_FLAG_SETS = 0
+_INVALID_FLAG_SETS = 0
 
 
 class Status(Enum):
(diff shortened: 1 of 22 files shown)


splitio/client/factory.py
```diff
@@ -684,7 +696,16 @@ def get_factory(api_key, **kwargs):
     _INSTANTIATED_FACTORIES.update([api_key])
     _INSTANTIATED_FACTORIES_LOCK.release()
 
-    config = sanitize_config(api_key, kwargs.get('config', {}))
+    config_raw = kwargs.get('config', {})
```
EmilianoSanchez, line 699: I am not sure if we need the global variables _TOTAL_FLAG_SETS, _INVALID_FLAG_SETS and _FLAG_SETS_LOCK to calculate total_flag_sets and invalid_flag_sets, because they are calculated per factory instance. Look if the following can work and help to simplify the code:
- remove lines 700 to 706
- replace `= _get_total_and_invalid_flag_sets()` with `= _get_total_and_invalid_flag_sets(cfg)`
- and implement `_get_total_and_invalid_flag_sets` as follows:
```
// Assuming that cfg has the same interface than config_raw
def _get_total_and_invalid_flag_sets(cfg):
    total_flag_sets = len(cfg.get('flagSetsFilter'))
    invalid_flag_sets = total_flag_sets - len(input_validator.validate_flag_sets(cfg.get('flagSetsFilter'), 'Telemetry Init'))
    return total_flag_sets, invalid_flag_sets
```

EmilianoSanchez (approved):
Bilal, please check the suggestion before merging. I think it can help to simplify the code, by removing global variables and thread locking.

splitio/client/factory.py
chillaq: Hi Emi, Every new factory instance will use the same static functions to create a factory instance, so the flag set calculation code could be  running for multiple factories at the same time. 
Its the same mechanism for the factories count per api key.

PR #479, opened 2023-10-24
Adding username option to Redis connections

craig-rueda:
# Python SDK

## What did you accomplish?

Adds optional `redisUsername` option to client config that allows downstream users to provide a user, along with the existing password which is important in the event that users would like to use a non-default username.

Should behave the same as the previous version as the defaults match the current kwarg defaults found in the Redis class (https://github.com/redis/redis-py/blob/master/redis/client.py#L201)

CC - @EmilianoSanchez


craig-rueda:
Bump - @EmilianoSanchez 😄

chillaq:
Hi Craig, Thanks for the suggestion, I don't see an issue with adding username into redis adapater, we will add next version.

I will let you know once it is released.

Thanks
Bilal

craig-rueda:
Cool! So you're going to make another PR, or would it work to merge this one? :)

Thanks

chillaq:
Hi @craig-rueda , I pushed the fix to development, it will be out with our next release.

Thanks
Bilal

PR #508, opened 2024-03-06
Sermver class

chillaq:
# Python SDK


splitio/models/grammar/matchers/semver.py
```diff
@@ -0,0 +1,158 @@
...
+        Class Initializer
+
+        :param version: raw version as read from splitChanges response.
+        :type version: str
+        """
+        self._major = 0
+        self._minor = 0
+        self._patch = 0
+        self._pre_release = []
+        self._is_stable = False
```
EmilianoSanchez, line 24: Should `_is_stable` be true by default, since `_pre_release` is an empty array by default?
Anyway, this is updated inside the `_parse` function

splitio/models/grammar/matchers/semver.py
```diff
@@ -0,0 +1,158 @@
...
+        Check if there is any metadata characters in self.version.
+
+        :returns: The semver string without the metadata
+        :rtype: str
+        """
+        index = version.find(self._METADATA_DELIMITER)
+        if index == -1:
+            return version
+
+        self._metadata = version[index:]
```
EmilianoSanchez, line 66: Check if it should throw an error if `_metadata` is an empty string

splitio/models/grammar/matchers/semver.py
```diff
@@ -0,0 +1,158 @@
...
+        """
+        without_metadata = self.remove_metadata_if_exists(version)
+
+        index = without_metadata.find(self._PRE_RELEASE_DELIMITER)
+        if index == -1:
+            self._is_stable = True
+        else:
+            pre_release_data = without_metadata[index+1:]
+            without_metadata = without_metadata[:index]
+            self._pre_release = pre_release_data.split(self._VALUE_DELIMITER)
```
EmilianoSanchez, line 51: I think here you should throw an error if some item in the `_pre_release` array is an empty string

EmilianoSanchez (approved):

splitio/models/grammar/matchers/semver.py
chillaq: will do

splitio/models/grammar/matchers/semver.py
chillaq: will do

splitio/models/grammar/matchers/semver.py
chillaq: yeah, it is updated in _parse.

PR #512, opened 2024-03-08
added semver between matcher

chillaq:
# Python SDK

### DIFF
--- a/splitio/models/grammar/matchers/semver.py
+++ b/splitio/models/grammar/matchers/semver.py
@@ -153,7 +153,7 @@ class EqualToSemverMatcher(Matcher):
         :param raw_matcher: raw matcher as fetched from splitChanges response.
         :type raw_matcher: dict
         """
-        self._data = raw_matcher['stringMatcherData']
+        self._data = raw_matcher.get('stringMatcherData')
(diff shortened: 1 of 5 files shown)


splitio/models/grammar/matchers/semver.py
```diff
@@ -277,3 +300,49 @@ def __str__(self):
...
+
+    def _build(self, raw_matcher):
+        """
+        Build a BetweenSemverMatcher.
+
+        :param raw_matcher: raw matcher as fetched from splitChanges response.
+        :type raw_matcher: dict
+        """
+        self._data = raw_matcher.get('betweenStringMatcherData')
+        self._semver_start = Semver(self._data['start']) if self._data.get('start') is not None else None
```
EmilianoSanchez, line 315: Should it be 
```
        self._semver_start = Semver.build(self._data['start'])
        self._semver_end = Semver.build(self._data['end'])
```
to avoid throwing an exception due to invalid semver string?

EmilianoSanchez (approved):

splitio/models/grammar/matchers/semver.py
chillaq: fixed

PR #513, opened 2024-03-08
added in list semver matcher

chillaq:
# Python SDK


splitio/models/grammar/matchers/semver.py
```diff
@@ -323,3 +346,51 @@ def __str__(self):
...
+        Build a InListSemverMatcher.
+
+        :param raw_matcher: raw matcher as fetched from splitChanges response.
+        :type raw_matcher: dict
+        """
+        self._data = raw_matcher.get('whitelistMatcherData')
+        if self._data is not None:
+            self._data = self._data.get('whitelist')
+
+        self._semver_list = [Semver(item) if item is not None else None for item in self._data] if self._data is not None else []
```
EmilianoSanchez, line 364: Maybe:
```
self._semver_list = [Semver.build(item) for item in self._data] if self._data is not None else []
```

EmilianoSanchez (approved):

splitio/models/grammar/matchers/semver.py
chillaq: fixed

PR #522, opened 2024-04-24
added spec query parameter

chillaq:
# Python SDK


splitio/api/commons.py
```diff
@@ -3,7 +3,7 @@
 
 _CACHE_CONTROL = 'Cache-Control'
 _CACHE_CONTROL_NO_CACHE = 'no-cache'
-
+_SPEC_VERSION = '1.1'
```
sanzmauro, line 6:
```
For consistency across our SDKs and to streamline reviews, issue tracking, and support, we aim to include a new file named spec at the root of each project. This file will define and allow us to update the value of the feature flags spec.
```

chillaq: will do

sanzmauro (approved):

PR #527, opened 2024-05-09
made some cleanup in code for semver

mmelograno:
# Python SDK


splitio/models/grammar/matchers/utils/__init__.py
sanzmauro, line 1: why do you need this file?

sanzmauro (approved):

PR #556, opened 2025-03-06
model and memory storage

chillaq:
# Python SDK

### DIFF
--- a/splitio/storage/__init__.py
+++ b/splitio/storage/__init__.py
@@ -355,3 +355,74 @@ class FlagSetsFilter(object):
             return False
 
         return any(self.flag_sets.intersection(flag_sets))
+    
+class RuleBasedSegmentsStorage(object, metaclass=abc.ABCMeta):
+    """SplitRule based segment storage interface implemented as an abstract class."""
+ 
+    @abc.abstractmethod
+    def get(self, segment_name):
+        """
+        Retrieve a rule based segment.
+
+        :param segment_name: Name of the segment to fetch.
+        :type segment_name: str
+
+        :rtype: str
+        """
+        pass
+
+    @abc.abstractmethod
+    def update(self, to_add, to_delete, new_change_number):
+        """
+        Update rule based segment..
+
+        :param to_add: List of rule based segment. to add
+        :type to_add: list[splitio.models.rule_based_segments.RuleBasedSegment]
+        :param to_delete: List of rule based segment. to delete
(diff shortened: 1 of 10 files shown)


splitio/storage/__init__.py
```diff
@@ -354,4 +354,75 @@ def intersect(self, flag_sets):
...
+        :rtype: int
+        """
+        pass
+
+    @abc.abstractmethod
+    def contains(self, segment_names):
+        """
+        Return whether the traffic type exists in at least one rule based segment in cache.
+
+        :param traffic_type_name: Traffic type to validate.
```
EmilianoSanchez, line 402: I think this doc comment corresponds to a different method, because traffic types are not involved in this method.

splitio/storage/inmemmory.py
```diff
@@ -107,6 +107,236 @@ def remove_flag_set(self, flag_sets, feature_flag_name, should_filter):
...
+        :param to_delete: List of rule based segment. to delete
+        :type to_delete: list[splitio.models.rule_based_segments.RuleBasedSegment]
+        :param new_change_number: New change number.
+        :type new_change_number: int
+        """
+        [self._put(add_segment) for add_segment in to_add]
+        [self._remove(delete_segment) for delete_segment in to_delete]
+        self._set_change_number(new_change_number)
+
+    def _put(self, rule_based_segment):
```
EmilianoSanchez, line 145: Should it be double underscored if it is a private method (`__`)?

splitio/storage/__init__.py
```diff
@@ -354,4 +354,75 @@ def intersect(self, flag_sets):
...
+
+        :param segment_name: Name of the segment to fetch.
+        :type segment_name: str
+
+        :rtype: str
+        """
+        pass
+
+    @abc.abstractmethod
+    def update(self, to_add, to_delete, new_change_number):
```
EmilianoSanchez, line 375: Probably not required in server-side, but the `update` method might return a boolean indicating whether the storage was actually modified or not. In client-side SDKs, this boolean is used as condition to emit the SDK_UPDATE event.

EmilianoSanchez (approved):
Check the comments before merging. Thanks Bilal!

splitio/storage/inmemmory.py
chillaq: nope, in python its one underscore, the double underscore are operators overload functions

splitio/storage/__init__.py
chillaq: will fix

PR #565, opened 2025-05-06
Rbs old spec localhost

chillaq:
# Python SDK


splitio/models/grammar/matchers/rule_based_segment.py
```diff
@@ -32,11 +32,14 @@ def _match(self, key, attributes=None, context=None):
         # Check if rbs segment has exclusions
         if context['ec'].segment_rbs_memberships.get(self._rbs_segment_name):
             return False
-                
-        for parsed_condition in context['ec'].segment_rbs_conditions.get(self._rbs_segment_name):
-            if parsed_condition.matches(key, attributes, context):
+        
+        for rbs_segment in context['ec'].excluded_rbs_segments:
+            if self._match_conditions(rbs_segment.conditions, key, attributes, context):
                 return True
```
sanzmauro, line 38: this should be return false

splitio/models/grammar/matchers/rule_based_segment.py
```diff
@@ -32,11 +32,14 @@ def _match(self, key, attributes=None, context=None):
         # Check if rbs segment has exclusions
         if context['ec'].segment_rbs_memberships.get(self._rbs_segment_name):
             return False
-                
-        for parsed_condition in context['ec'].segment_rbs_conditions.get(self._rbs_segment_name):
-            if parsed_condition.matches(key, attributes, context):
+        
+        for rbs_segment in context['ec'].excluded_rbs_segments:
+            if self._match_conditions(rbs_segment.conditions, key, attributes, context):
```
sanzmauro, line 37: we need to do something like this 
```java
    private boolean matchExcludedSegments(List<ExcludedSegments> excludedSegments, Object matchValue, String bucketingKey,
                                          Map<String, Object> attributes, EvaluationContext evaluationContext) {
        for (ExcludedSegments excludedSegment: excludedSegments) {
            if (excludedSegment.isStandard() && evaluationContext.getSegmentCache().isInSegment(excludedSegment.name, (String) matchValue)) {
                return true;
            }

            if (excludedSegment.isRuleBased()) {
                RuleBasedSegmentMatcher excludedRbsMatcher = new RuleBasedSegmentMatcher(excludedSegment.name);
                if (excludedRbsMatcher.match(matchValue, bucketingKey, attributes, evaluationContext)) {
                    return true;
                }
            }
        }

        return  false;
    }
```

splitio/models/grammar/matchers/rule_based_segment.py
```diff
@@ -32,11 +32,14 @@ def _match(self, key, attributes=None, context=None):
...
                 return True
-                
+        
+        if self._match_conditions(context['ec'].segment_rbs_conditions.get(self._rbs_segment_name), key, attributes, context):
+            return True
+                                
         return False
```
sanzmauro, line 43:
```suggestion
        return self._match_conditions(context['ec'].segment_rbs_conditions.get(self._rbs_segment_name), key, attributes, context)
```

sanzmauro (changes requested):

splitio/models/grammar/matchers/rule_based_segment.py
chillaq: will do

splitio/models/grammar/matchers/rule_based_segment.py
chillaq: The function already return false. We don't pass the storage objects to matchers in Python, so async and thread base will use the same calls.

splitio/models/grammar/matchers/rule_based_segment.py
chillaq: The function already return false. We don't pass the storage objects to matchers in Python, so async and thread base will use the same calls.

splitio/models/grammar/matchers/rule_based_segment.py
```diff
@@ -29,20 +30,45 @@ def _match(self, key, attributes=None, context=None):
         if self._rbs_segment_name == None:
             return False
         
-        # Check if rbs segment has exclusions
-        if context['ec'].segment_rbs_memberships.get(self._rbs_segment_name):
+        rb_segment = context['ec'].rbs_segments.get(self._rbs_segment_name)
```
sanzmauro, line 33: could `rb_segment` be null?

splitio/models/grammar/matchers/rule_based_segment.py
```diff
@@ -29,20 +30,45 @@ def _match(self, key, attributes=None, context=None):
...
+                return True
+            
+        return False
+            
+    def _match_dep_rb_segments(self, excluded_rb_segments, key, attributes, context):
+        for excluded_rb_segment in excluded_rb_segments:
+            if excluded_rb_segment.type == SegmentType.STANDARD:
+                if context['ec'].segment_memberships[excluded_rb_segment.name]:
+                    return True
+            else:
```
sanzmauro, line 63: mmm, what happen if `type` is equals to `large` ?

splitio/models/grammar/matchers/rule_based_segment.py
```diff
@@ -29,20 +30,45 @@ def _match(self, key, attributes=None, context=None):
...
+                excluded_segment = context['ec'].rbs_segments.get(excluded_rb_segment.name)
+                if key in excluded_segment.excluded.get_excluded_keys():
+                    return False
```
sanzmauro, line 66: why this return `false`?

splitio/models/grammar/matchers/rule_based_segment.py
```diff
@@ -29,20 +30,45 @@ def _match(self, key, attributes=None, context=None):
...
+
+                if self._match_dep_rb_segments(excluded_segment.excluded.get_excluded_segments(), key, attributes, context):
+                    return True
+                    
+                if self._match_conditions(excluded_segment.conditions, key, attributes, context):
+                    return True 
+
+        return False
```
sanzmauro, line 74:
```suggestion
        return self._match_conditions(excluded_segment.conditions, key, attributes, context)
```

sanzmauro (commented):
i left some comments

splitio/models/grammar/matchers/rule_based_segment.py
chillaq: It matches to false if excluded keys contain the matching key

splitio/models/grammar/matchers/rule_based_segment.py
chillaq: I thought we don't support it yet

tests/engine/files/rule_base_segments.json
```diff
@@ -0,0 +1,62 @@
+{"ff": {"d": [], "t": -1, "s": -1},
+"rbs": {"t": -1, "s": -1, "d":
```
EmilianoSanchez, line 2: I think this doesn't change or break the test, but `t` should be greater than `s` if there is a diff in definitions.  Maybe this helps to detect some bug. Some in other response mocks.

EmilianoSanchez (approved):

tests/engine/files/rule_base_segments.json
chillaq: Yeah, I do have proper sync tests with `s` and `t` are set correctly, in this test though, it is only for evaluation and matcher tests

PR #608, opened 2026-01-07
added models, events config data and events metadata

chillaq:
# Python SDK

### DIFF
--- a/splitio/events/events_manager_config.py
+++ b/splitio/events/events_manager_config.py
@@ -1,0 +1,124 @@
+"""Events Manager Configuration."""
+from splitio.models.events import SdkEvent, SdkInternalEvent
+
+class EventsManagerConfig(object):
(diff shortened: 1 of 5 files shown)


splitio/events/events_manager_config.py
```diff
@@ -0,0 +1,124 @@
...
+    
+    @property
+    def execution_limits(self):
+        """Return execution_limits dict"""
+        return self._execution_limits
+    
+    @property
+    def prerequisites(self):
+        """Return require all dict"""
+        return self._prerequisites
```
gthea: Think this one's duplicated

gthea (approved):

splitio/events/events_manager_config.py
chillaq: will remove

PR #641, opened 2026-06-23
Bug fixes and polishing

EmilianoSanchez:
# Python SDK


splitio/api/client.py
```diff
@@ -108,7 +108,7 @@ def __init__(self, timeout=None, sdk_url=None, events_url=None, auth_url=None, t
         :type telemetry_url: str
         """
         _LOGGER.debug("Initializing httpclient")
-        self._timeout = timeout/1000 if timeout else None # Convert ms to seconds.
+        self._timeout = timeout if timeout else None
```
chillaq: This is a breaking change, need management approval

splitio/client/config.py
```diff
@@ -18,7 +18,7 @@ class AuthenticateScheme(Enum):
 
 DEFAULT_CONFIG = {
     'operationMode': 'standalone',
-    'connectionTimeout': 1500,
+    'connectionTimeout': 15,
```
chillaq: Breaking change

splitio/api/client.py
EmilianoSanchez: Yes. Change reverted

splitio/client/config.py
EmilianoSanchez: Change reverted

chillaq (approved):

